PR #42, opened 2022-09-09
Add more SIP editors

rekmarks:
Adds the following SIP editors by adding their GitHub handles to the `CODEOWNERS` file:
- @hmalik88 
- @Montoya 
- @rekmarks (yours truly)


Mrtenz:
You should add them to [SIP-1](https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-1.md#sip-editors) too.

ritave (approved):

PR #45, opened 2022-09-19
Setup Jekyll for site generation

### LINKED ISSUE #44: Add a validation rule that verifies Table of Contents
Some SIPs add a Table of Contents, which may become desynced from the paragraphs. We should add a rule that verifies if they still match.

The algorithm for converting headers into `[Link Slugs](#link-slugs)` is not simple and may require some time to figure out all the edges cases.

FrederikBolding:
Sets up Jekyll for site generation. Based on the set up used for EIPs and CAIPs (pending review).

Removes existing tables of content as those can be automatically generated.

fixes #44


_includes/siptable.html
```diff
@@ -0,0 +1,36 @@
...
+    }
+  </style>
+  {% for status in site.data.statuses %}
+    {% assign sips = include.sips|where:"status",status|sort:"sip" %}
+    {% assign count = sips|size %}
+    {% if count > 0 %}
+      <h2 id="{{status|slugify}}">{{status}}</h2>
+      <table class="siptable">
+        <thead>
+          {% if status == "Last Call" %}
```
ritave, line 17: We don't have a `Last Call` status

_layouts/sip.html
```diff
@@ -0,0 +1,106 @@
...
+      <th>Created</th>
+      <td>{{ page.created | xml_escape }}</td>
+    </tr>
+    {% if page.updated != undefined %}
+    <tr>
+      <th>Updated</th>
+      <td>{{ page.updated | xml_escape }}</td>
+    </tr>
+    {% endif %}
+    {% if page.requires != undefined %}
```
ritave, line 54: Multiple properties in preamble that don't exist in SIPs

_layouts/sip.html
```diff
@@ -0,0 +1,106 @@
...
+
+  <div class="toc">
+    <h2>Table of Contents</h2>
+    {% include toc.html html=content sanitize=true h_max=3 %}
+  </div>
+
+  {% include anchor_headings.html html=content anchorClass="anchor-link" beforeHeading=true %}
+
+  <h2>Citation</h2>
+  <p>Please cite this document as:</p>
```
ritave, line 76: Hah, quite a big ego for us to be thinking someone will be citing Snaps

ritave (changes requested):
Minor problems:
1. The ones in the comments
2. The newlines at the end of the files are inconsistent. Please add them everywhere.
3. Might as well add a link in a README.md to the Github Pages

Questions:
1.  Is it possible to hide Jekyll files to a sub-folder so that they don't pollute root?

_layouts/sip.html
FrederikBolding: Just included from the EIP/CAIP template. I can remove it if you prefer

_layouts/sip.html
ritave: Sure, sounds useful

_layouts/sip.html
ritave: No opinion, up to you.

ritave (approved):

Issue #62, opened 2022-10-20
SIP-5: Creating JSON Web Tokens (JWTs)


Vid201:
A place to discuss SIP-5: Creating JSON Web Tokens (JWTs). The SIP proposes a new signing method - JWTs (one of the most used data standards on the internet).

ritave:
Closing in favour of #64 discussion

PR #65, opened 2022-10-20
Added SIP-5

Vid201:
SIP-5: Creating JSON Web Tokens (JWTs)

https://github.com/MetaMask/SIPs/issues/62

### DIFF
--- a/SIPS/sip-5.md
+++ b/SIPS/sip-5.md
@@ -65,16 +65,20 @@ interface JWTPayload {
   iat?: number;
   [claims: string]: any;
 }
-
-type HashingFunction = (data: string | Uint32Array) => string | Uint32Array;
 ```
 
 ### JWT Signing Implementation
 
 This implementation is modeled after implementations from libraries [did-jwt](https://github.com/decentralized-identity/did-jwt) and [jose](https://github.com/panva/jose).
 
+All hash functions that are implemented in the library [ethereum-cryptography](https://github.com/ethereum/js-ethereum-cryptography) are supported, e.g., SHA256, SHA512, and keccak256.
+
 ```typescript
 import * as u8a from "uint8arrays";
+const { sha256 } = require("ethereum-cryptography/sha256");
+const { keccak256 } = require("ethereum-cryptography/keccak");
+// import other hash functions
+const { utf8ToBytes } = require("ethereum-cryptography/utils");
 
 function hexToBytes(s: string): Uint8Array {
   const input = s.startsWith("0x") ? s.substring(2) : s;
@@ -97,7 +101,7 @@ export async function createJWT(
   header: Partial<JWTHeader> = {},
   payload: Partial<JWTPayload> = {},
   address: string,
-  hashingFunction: HashingFucntion
+  hashFunction: string
 ): Promise<string> {
   if (!header.typ) header.typ = "JWT";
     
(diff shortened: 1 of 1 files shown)


SIPS/sip-5.md
```diff
@@ -0,0 +1,176 @@
...
+    
+  const encodedPayload = (typeof payload === "string") ? payload : encodeSection(payload);
+  const signingInput: string = [encodeSection(header), encodedPayload].join(
+    "."
+  );
+
+  const hash = hashingFunction(signingInput);
+  const bytes = hexToBytes(hash);
+
+  let signature = /* eth_sign MetaMask RPC method that creates the signature (showing the user header and payload that will be signed) */;
```
Mrtenz, line 112: Do you have any more details on this? What's the actual payload to sign?

SIPS/sip-5.md
```diff
@@ -0,0 +1,176 @@
...
+            "typ": "JWT"
+        },
+      payload: 
+        {
+            "iss": "Company",
+            "sub": "Alice",
+            "role": "employee"
+        },
+      address: '0x12345...',
+      hashingFunction: sha256 // a custom hashing function can be used instead
```
Mrtenz, line 143: This won't work unfortunately. JSON-RPC,  as implied by the name, use JSON. JavaScript functions aren't JSON serialisable.

SIPS/sip-5.md
Vid201: Yes, sorry for the ambiguity. It is `bytes` from line 110.

SIPS/sip-5.md
Vid201: Ah, true. Then the way to go is to pass the hashing function name as a parameter (enum style, e.g., `hashingFunction: 'sha256'`) and use the correct hashing function accordingly:

```typescript
switch (hashingFunction) {
    case 'sha256':
        // sha256 hash
    case 'keccak256':
        // keccak256 hash
    ...
}
```

I think all hashing functions from the library 'ethereum-cryptography' (which is used in `eth-sig-util`) can be available: https://github.com/ethereum/js-ethereum-cryptography#hashes-sha256-keccak-256-ripemd160-blake2b

SIPS/sip-5.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Motivation
+
+Snaps were introduced to extend the functionalities of MetaMask to support more use cases and paradigms. MetaMask currently provides several RPC methods for digitally signing data (described [here](https://docs.metamask.io/guide/signing-data.html)), such as personal_sign and signTypedData. All methods add the prefix "\x19Ethereum Signed Message:\n" to the data, as described in [EIP 191](https://eips.ethereum.org/EIPS/eip-191) and [EIP 712](https://eips.ethereum.org/EIPS/eip-712), which prevents several attack vectors, most notably impersonating transactions. But this also prevents the ability to produce pure signatures over the data, which is needed for other internet data standards (such as JWTs). Actually, a pure signature is possible only with eth_sign, but that method was deprecated and advised not to use. Therefore, there is no safe way to create JWTs containing signatures signed by MetaMask accounts. This SIP proposes a new RPC method for Snaps that enables the creation of signed JWTs.
```
ritave, line 18: Can you explain the rationale of why the snap itself can't calculate JWT cookies using `snap_getBip44Entropy`?

The above function gives you the raw, extended, private key that is used inside MetaMask itself. With that entropy you can do calculations inside of the snap, without needing additional support from the extension.

Vid201: It is possible. There are two ways to do it without additional support:

1. Using the RPC method `eth_sign`, but this method is deprecated, and the user can see a red notice when signing.
2. Exporting private key with `snap_getBip44Entropy`, but then the user must confirm that it allows Snap to export private key (when installing Snap) from "main" MetaMask to the Snap. But this would look suspicious to users and prevent them from using the Snap. We also need to sign with the same private key as used on the Ethereum, which means Snap has complete access to the funds on the mainnet.

Understandably, MetaMask and Snap JSON-RPC API cannot support all signing methods, but JWTs are one of the most used standards on the internet, and many Snaps could reuse this functionality.

ritave (approved):
From the perspective of the Snaps team, this functionality should be handled by the `snap_getBip44Entropy`.

On the other hand, this SIP has technical merits and so I will approve and merge this proposal on the editor's grounds. I'd like to see if any discussions breaks up concerning this change. If the community supports this change, it will be considered!

Keep in mind, that if no support from the community occurs, there's a high probability the Snaps team will not spend time to implement this change.

PR #68, opened 2022-10-27
Add SIP-6: Deterministic Snap-specific entropy

Mrtenz:
This adds a new SIP, SIP-6: Deterministic Snap-specific entropy. It describes a way for Snaps to get entropy which cannot be recreated by other Snaps.

### DIFF
--- a/SIPS/sip-6.md
+++ b/SIPS/sip-6.md
@@ -43,6 +43,9 @@ path. This derivation path MUST start with the magic value `0xd36e6170` (`139974
 following eight indices are based on a hash of the Snap ID, with an optional salt. The hashing algorithm of choice is
 `keccak256`.
 
+To prevent other Snaps from getting the same entropy, Snaps MUST NOT be able to manually derive using the magic value
+`0xd36e6170`, i.e., through the `snap_getBip32Entropy` JSON-RPC method.
+
 The hash of the Snap ID is calculated as follows:
 
 ```typescript
--- a/SIPS/sip-6.md
+++ b/SIPS/sip-6.md
@@ -2,7 +2,7 @@
 sip: 6
 title: Deterministic Snap-specific entropy
 status: Draft
-discussions-to: (Http/Https URL)
+discussions-to: https://github.com/MetaMask/SIPs/discussions/69
 author: Maarten Zuidhoorn (@Mrtenz)
 created: 2022-10-27
 ---
--- a/SIPS/sip-6.md
+++ b/SIPS/sip-6.md
@@ -145,6 +145,7 @@ const getDerivationPath = (indices: number[]) => {
  * @param snapId - The ID of the Snap.
  * @param salt - An optional salt to use in the derivation. If not provided, an
  * empty string is used.
+ * @returns The derived entropy.
  */
 const getEntropy = async (
   mnemonicPhrase: string,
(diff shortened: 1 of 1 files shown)


SIPS/sip-6.md
```diff
@@ -0,0 +1,184 @@
...
+```typescript
+const hash = keccak256(snapId + keccak256(salt));
+```
+
+If the salt is not provided, an empty string MUST be used instead:
+
+```typescript
+const hash = keccak256(snapId + keccak256(''));
+```
+
```
Montoya, line 57: The developer of the snap can decide whether to use a salt or not? What is the purpose of it? Why would they add a salt, or choose not to?

SIPS/sip-6.md
```diff
@@ -0,0 +1,184 @@
...
+unable to get the same entropy.
+
+Snaps can optionally specify a salt in order to generate different entropy for different use cases, for example, some
+entropy to derive new private keys from, and some other entropy to encrypt some data.
+
+The entropy can be accessed from within a Snap, using the `snap_getEntropy` JSON-RPC method.
+
+## Motivation
+
+Before this SIP, Snaps did not have a way to get some kind of deterministic entropy, without storing it on the disk. If
```
FrederikBolding, line 23: If possible it might be interesting to use this concept for state encryption as well.

SIPS/sip-6.md
```diff
@@ -0,0 +1,184 @@
...
+```typescript
+const hash = keccak256(snapId + keccak256(''));
+```
+
+The hash is then split into eight big endian `uint32 | 0x80000000` integers. The resulting derivation path is a
+combination of the magic value, and the eight integers:
+
+```typescript
+const computedDerivationPath = getUin32Array(hash).map((index) => (index | 0x80000000) >>> 0);
+const derivationPath = [0xd36e6170, ...computedDerivationPath];
```
FrederikBolding, line 63: Should this path be blocked in `getBip32Entropy`?

SIPS/sip-6.md
```diff
@@ -0,0 +1,184 @@
...
+[BIP 32](https://github.com/bitcoin/bips/blob/6545b81022212a9f1c814f6ce1673e84bc02c910/bip-0032.mediawiki), using a root
+node created as per
+[BIP 39](https://github.com/bitcoin/bips/blob/6545b81022212a9f1c814f6ce1673e84bc02c910/bip-0039.mediawiki).
+
+### `snap_getEntropy` JSON-RPC method
+
+The `snap_getEntropy` JSON-RPC method is used to get the entropy for a Snap. It takes a single parameter, which is an
+object with the following properties:
+
+- `salt` (optional, `string`): A salt to use when deriving the entropy. If provided, this MUST be interpreted as a UTF-8
```
FrederikBolding, line 83: How would a snap use `snap_getEntropy` for different chains? Would they need different salts or should we have a built-in coinType or similar param?

SIPS/sip-6.md
Mrtenz: It should. I've added a section to the SIP to clarify.

SIPS/sip-6.md
Mrtenz: `snap_getEntropy` can be used to get general entropy, so it's not specific to BIP-32/44 key derivation. Snaps wanting to derive keys for different chains can use the entropy as seed, and derive BIP-44 nodes from it.

hmalik88 (approved):

PR #70, opened 2022-10-27
Add SIP-7: Snaps UI

Mrtenz:
This adds a new SIP, SIP-7: Snaps UI. It describes a way for Snaps create a custom UI.

https://github.com/MetaMask/snaps-monorepo/issues/707

### DIFF
--- a/SIPS/sip-7.md
+++ b/SIPS/sip-7.md
@@ -274,7 +274,7 @@ export const onRpcRequest = async () => {
     method: 'snap_updateInterface',
     params: {
       id,
-      json: {
+      ui: {
         type: 'panel',
         children: [
           /* Some components based on the `data`. */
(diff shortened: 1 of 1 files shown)


SIPS/sip-7.md
```diff
@@ -0,0 +1,308 @@
...
+interface OnTransactionReturn {
+  insights: Component;
+}
+
+type OnTransaction = (args: OnTransactionArgs) => Promise<OnTransactionReturn>;
+```
+
+### Example
+
+The Snaps UI may also be used similarly to `snap_dialog`, but SHALL NOT be supported by `snap_dialog` itself. Instead, it may be used as so:
```
Montoya, line 254: I'm confused about this part. snap_dialog doesn't use this UI API?

PR #82, opened 2022-11-10
Add version number to SIP-6

Mrtenz:
As discussed internally, the `snap_getEntropy` method should have a version number, so we can change the behaviour in the future without the need for a new method.

For SIP-6, it's hardcoded to the number `1`.

### DIFF
--- a/SIPS/sip-6.md
+++ b/SIPS/sip-6.md
@@ -80,9 +80,10 @@ node created as per
 
 ### `snap_getEntropy` JSON-RPC method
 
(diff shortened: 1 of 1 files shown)


ritave (changes requested):
Requires further discussion

ritave:
Why not add the version when we actually break it.
If the version field doesn't exist -> it's version 1. Otherwise, use the version specified.

ritave (approved):
Approving since it's already implemented.

PR #83, opened 2022-11-15
Moving SIP-7 to object based exports

ritave:


FrederikBolding (commented):
In my opinion the SIP also needs to specify how this should work behind the scenes, which is likely using RPC methods.

It was always the plan to add something nice on top of the suggested RPC methods to make using it simpler. But to make the SIP explicit, we started of with specifying the base layer of the UI system.

SIPS/sip-7.md
```diff
@@ -2,307 +2,205 @@
 sip: 7
 title: Snaps UI
 status: Draft
-discussions-to: https://github.com/MetaMask/SIPs/discussions/71
-author: Frederik Bolding (@FrederikBolding), Maarten Zuidhoorn (@Mrtenz)
+discussions-to: https://github.com/wallet/SIPs/discussions/71
```
FrederikBolding, line 5:
```suggestion
discussions-to: https://github.com/MetaMask/SIPs/discussions/71
```

ritave:
> In my opinion the SIP also needs to specify how this should work behind the scenes, which is likely using RPC methods.
SIPs are not really a place to talk about implementation details. See SIP-2 for example, it only discusses end-user facing parts of the contract - what's the API, how are the keyrings executed, what is the expectations, etc.

MetaMask implementation details are not interesting to end-developers who read those SIPs.

I'll look into other comments tomorrow.

SIPS/sip-7.md
```diff
@@ -2,307 +2,205 @@
...
-```tsx
-interface SnapResolveInterfaceRequest {
-  method: 'snap_resolveInterface';
-  params: {
-    id: string;
-    value: Json;
-  };
-}
+```typescript
+module.exports.ui = new MyUI();
```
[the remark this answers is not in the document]
ritave: I'll write more in the SIP.

TL;DR;

```typescript
const dialogs = /* ... */;

onChange = ({ surface, id, action: { id: inputId, value } }) => {
  assert(surface === "dialog");
  assert(dialogs.get(id)?.inputs.get(inputId) !== undefined)
};
```

It's all based on the proposed API I did a while ago
https://gist.github.com/ritave/936013d84a0111ae27046e940915c5cc

SIPS/sip-7.md
```diff
@@ -2,307 +2,205 @@
...
-
-#### `snap_showInterface`
-
-The `snap_showInterface` method lets Snaps show a user interface arbitrarily. The params MUST be an object, with the following properties:
+## Core Components
 
-- `ui` (`Component`): The custom UI to show.
+> Defining a complete set of components used in the wallet is outside of scope for this SIP.
+>
+> Only a core set of components is defined, delegating definition of a complete set of useful components to other SIPs.
```
Mrtenz, line 76: Why? We had already defined a good set of initial components, other SIPs can expand on this.

SIPS/sip-7.md
```diff
@@ -2,307 +2,205 @@
...
-
-interface ButtonComponent extends BaseComponent {
-  type: 'button';
-  variant?: 'primary' | 'secondary';
-  text: string;
-  name?: string;
+```typescript
+interface Parent extends Node {
+  children: Node[];
 }
```
Mrtenz, line 61: I don't think we need a separate `Parent` interface. Nodes that can have children can just define that in their spec.

SIPS/sip-7.md
```diff
@@ -2,307 +2,205 @@
...
+## Definitions
 
-```tsx
-type Json =
-  | null
-  | boolean
-  | number
-  | string
-  | Json[]
-  | { [prop: string]: Json };
```
Mrtenz, line 37: Why did you remove this? I think it's still important to clarify what we mean by "JSON-serializable".

SIPS/sip-7.md
Mrtenz, line 104: I don't understand the need for a class here. This could be implemented using an object, as the class likely won't hold any meaningful state.

SIPS/sip-7.md
```diff
@@ -2,307 +2,205 @@
...
-Snaps that require user input in order to continue execution, i.e., checking if a user pressed the confirm or deny button, can use the `snap_readInterface` method to read the resolved value.
+The interface for the exposed object is as follows
 
-The params MUST be an object, with the following properties:
+```typescript
 
-- `id` (`string`): The ID of the custom UI to read.
+// Specific Surfaces definitions are out of scope for this SIP.
+type Surface = string;
+type InputAction = { id?: string; value?: JSON };
```
Mrtenz, line 113:
```suggestion
type InputAction = { id?: string; value?: Json };
```

SIPS/sip-7.md
```diff
@@ -2,307 +2,205 @@
...
+type OutputAction = { ui?: UI; action?: "close" };
 
-This method returns the resolved value. If the interface is not resolved yet, this method MUST wait until the interface is resolved, before returning a value.
+type OnChangeArguments =
```
Mrtenz, line 116: Did you forget something here?

Mrtenz (changes requested):

SIPS/sip-7.md
ritave: I'll write a second SIP in this PR as well.

This will allow this SIP to become Final, specifying the API in general, and the second to be considered Living, and be continually updated with new components as we go.

SIPS/sip-7.md
```diff
@@ -2,307 +2,205 @@
...
-interface DividerComponent extends BaseComponent {
-  type: 'divider';
-}
+### Literal
 
-interface SpacerComponent extends BaseComponent {
-  type: 'spacer';
+```typescript
+interface Literal extends Node {
+  value: any;
```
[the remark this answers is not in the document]
ritave: It's an abstract class that is not supposed to be used by snaps. The concrete examples will override the value to a more concrete type.

The whole structure was inspired by [Unist](https://github.com/syntax-tree/unist)

SIPS/sip-7.md
ritave: The whole structure was inspired by [Unist](https://github.com/syntax-tree/unist). Having defined Parent and Literal brings common fields so that automatic tools can operate on the whole UI structure without needing to understand each Node separately.

An example of such tool is [unist-util-visit](https://github.com/syntax-tree/unist-util-visit#visittree-test-visitor-reverse), which is extremely useful and [used extensively](https://github.com/MetaMask/SIPs/blob/60c212d1da607005bef989fd364565f21c66f111/tools/validate/src/rules/bad-link.ts#L86) in SIP validation tool and other local parser experiments I've build when playing with UI.

SIPS/sip-7.md
ritave: IMHO it shouldn't be part of this SIP to define what a JSON-serializable value is, because we used that term in multiple other places.

SIPS/sip-7.md
ritave: No need for a class. It's to showcase that it should be an *instance* or an object, rather than prototype of a class.

Same explanation happens in SIP-2

SIPS/sip-7.md
ritave: It's a draft, unfinished :)

ritave:
On hold until we get Snaps to MetaMask main extension

PR #98, opened 2023-06-30
Add SIP-10: snap_getLocale

FrederikBolding:
Adds SIP-10, which presents a new RPC method: `snap_getLocale`.

Fixes https://github.com/MetaMask/snaps/issues/1509

### DIFF
--- a/SIPS/sip-10.md
+++ b/SIPS/sip-10.md
@@ -2,7 +2,7 @@
 sip: 10
 title: wallet_getLocale
 status: Draft
-discussions-to (*optional): (Http/Https URL)
+discussions-to: https://github.com/MetaMask/SIPs/discussions/99
 author: Frederik Bolding (@FrederikBolding)
 created: 2023-06-30
 ---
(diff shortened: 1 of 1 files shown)


Mrtenz (dismissed):
This looks good to me, but I feel like we should name this `snap_getLocale` since it will be only available to snaps.

SIPS/sip-10.md
```diff
@@ -0,0 +1,64 @@
...
+}
+```
+
+```tsx
+/**
+ * Method hook for accessing the user locale from the PreferencesController.
+ *
+ * @returns The user selected locale.
+ */
+function getLocale(): string {
```
hmalik88, line 57: Is the locale string formatted a certain way? Wondering if we need to narrow the type.

hmalik88: I believe we're using IETF BCP 47. https://learn.microsoft.com/en-us/globalization/locale/standard-locale-names

hmalik88 (approved):

Mrtenz (approved):

Montoya (approved):

PR #102, opened 2023-07-16
Add SIP-11 : Transaction Insights V2

hmalik88:
This SIP updates the transaction insight API.

### DIFF
--- a/SIPS/sip-11.md
+++ b/SIPS/sip-11.md
@@ -1,5 +1,5 @@
 ---
-sip: 10
+sip: 11
 title: Transaction Insights V2 - Transaction Warnings
 status: Draft
 discussions-to: https://github.com/MetaMask/SIPs/discussions/101
--- a/SIPS/sip-10.md
+++ b/SIPS/sip-10.md
@@ -1,0 +1,70 @@
+---
+sip: 10
+title: Transaction Insights V2 - Transaction Warnings
+status: Draft
+discussions-to: https://github.com/MetaMask/SIPs/discussions/101
+author: Hassan Malik (@hmalik88)
+created: 2023-07-16
+---
+
+## Abstract
+
+This SIP proposes a way for snaps to provide extra friction for executing potentially dangerous transactions within MetaMask.
+
+## Motivation
+
+One of the biggest issues with wallet users is a loss of funds through executing some sketchy or seemingly innocuous transaction. With the benefit of the transaction insights API, a snap can already run analysis on an unsigned transaction paylod. Why not allow for the snap to provide extra friction to the user if it determines that a transaction is malicious? Users often click through things without reading, by allowing for a warning we can add friction at points that we really think a user should have a second glance.
+
+This warning could be injected at any point where there are insights provided.
+
+## Specification
+
+> Formal specifications are written in Typescript. Usage of `CAIP-N` specifications, where `N` is a number, are references to [Chain Agnostic Improvement Proposals](https://github.com/ChainAgnostic/CAIPs).
+
+Please see SIP-3 for more information on the original transaction insights API.
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Implementation
+
+Any snap that wishes to provide transaction insight features **MUST** implement the following API:
+
+```typescript
+import { OnTransactionHandler } from "@metamask/snap-types";
+
+export const onTransaction: OnTransactionHandler = async ({
+  transaction,
+  chainId,
+}) => {
+  const insights = /* Get insights */;
+  const willFail = () => /* Return boolean */;
+  return willFail(insights) ? { insights, warning: 'Some warning message' } : { insights };
+};
+```
+
+The interface for the return value of an `onTransaction` export is:
+
+```typescript
+interface OnTransactionReturn {
+  insights: Record<string, Json>;
+  warning?: string;
+}
+```
+
+### MetaMask Integration
+
+The `insights` object returned by the snap will be displayed alongside the confirmation for the `transaction` that `onTransaction` was called with.
+Keys and values will be displayed in the order received, with each key rendered as a title and each value rendered as follows:
+
+- If the value is an array or an object, it will be rendered as text after being converted to a string.
+- If the value is neither an array nor an object, it will be rendered directly as text.
+
+The `warning` key's value will be displayed in a modal alongside any warnings from other transaction insight snaps. The modal will require a checkbox to be checked before the user can continue with the transaction.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE).


SIPS/sip-11.md
```diff
@@ -0,0 +1,62 @@
...
+  return willFail(insights) ? { insights, warning: 'Some warning message' } : { insights };
+};
+```
+
+The interface for the return value of an `onTransaction` export is:
+
+```typescript
+interface OnTransactionReturn {
+  insights: Record<string, Json>;
+  warning?: string;
```
FrederikBolding, line 52: I thought this was supposed to be a flag or an enum value instead of a string?

Are we not just showing the insights content again in the modal?

cc @eriknson

SIPS/sip-11.md
```diff
@@ -0,0 +1,62 @@
...
+  const willFail = () => /* Return boolean */;
+  return willFail(insights) ? { insights, warning: 'Some warning message' } : { insights };
+};
+```
+
+The interface for the return value of an `onTransaction` export is:
+
+```typescript
+interface OnTransactionReturn {
+  insights: Record<string, Json>;
```
FrederikBolding, line 51: We don't support this API anymore, we only support custom UI returned.

FrederikBolding (dismissed):

SIPS/sip-11.md
hmalik88: Oh maybe? I was just going off the designs. Haven't had a chance to speak to Erik yet.

SIPS/sip-11.md
hmalik88: TIL, I'll update.

SIPS/sip-11.md
eriknson: Oh ye, it's the same insight. And it'll only populate the modal if it's in critical / warning state (boolean)

FrederikBolding: I think we should have this be an enum instead of a boolean value, for now we can just leave the enum with two states: "none" or "critical". This makes it easy to expand to add more states if we ever want to.

Montoya (dismissed):
Will wait for the feedback from Frederik to be addressed. 

I think the SIP should include mentioning that if the snap wishes to warn the user about potentially risky transactions then it will receive the transaction object in every transaction, rather than the current approach where the snap is only triggered if the tab is viewed. That part is not clear to me. How does MetaMask ensure this? Is it a caveat on the endowment?

hmalik88:
@Montoya I don't think the extension works like that. There are some transaction flows where the transaction object is modified and others where it isn't, so there isn't a single source of truth you can rely on for transaction parameters afaik. We have to write the code around different parts of the extension to surface this modal. The snap will provide insights and an opinion if the insights are critical or safe regardless.

Montoya:
OK but right now with transaction insights, if the user does not view your snap in the tab, then the export is never called, correct?

hmalik88:
That's correct. OK, I think I understand the question you were getting at and the intention behind it perhaps. I don't think it makes sense to have a caveat for surfacing the insight to the modal because it is a dynamic thing. The snap doesn't know if the modal population will be needed until it does, so the caveat becomes extraneous. What we would be able to do is fire the snap requests outside of clicking the tab with this v2 change. Will all confirmations have tabs now for snaps?

Montoya (approved):
LGTM

SIPS/sip-11.md
```diff
@@ -0,0 +1,75 @@
...
+
+```typescript
+enum SeverityLevel {
+  None = 'none',
+  Critical = 'critical',
+}
+
+interface OnTransactionResponse {
+  content: Component | null;
+  severity: SeverityLevel;
```
FrederikBolding, line 61: Should be optional imo. Where not returned means "none", that way we don't have a breaking change in the API either.

hmalik88: Hmm ok good point, I'll change.

SIPS/sip-11.md
```diff
@@ -0,0 +1,76 @@
...
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Implementation
+
+Any snap that wishes to provide transaction insight features **MUST** implement the following API:
+
+```typescript
+import { OnTransactionHandler } from "@metamask/snap-types";
+
+export const onTransaction: OnTransactionHandler = async ({
```
Mrtenz, line 39: This is an example, not an API description.

Mrtenz: The example is a bit confusing too.

SIPS/sip-11.md
```diff
@@ -0,0 +1,76 @@
...
+created: 2023-07-16
+---
+
+## Abstract
+
+This SIP proposes a way for snaps to provide extra friction for executing potentially dangerous transactions within MetaMask.
+
+## Motivation
+
+One of the biggest issues with wallet users is a loss of funds through executing some sketchy or seemingly innocuous transaction. With the benefit of the transaction insights API, a snap can already run analysis on an unsigned transaction paylod. Why not allow for the snap to provide extra friction to the user if it determines that a transaction is malicious? Users often click through things without reading, by allowing for a warning we can add friction at points that we really think a user should have a second glance.
```
Mrtenz, line 16:
```suggestion
One of the biggest issues with wallet users is a loss of funds through executing some sketchy or seemingly innocuous transaction. With the benefit of the transaction insights API, a snap can already run analysis on an unsigned transaction payload. Why not allow for the snap to provide extra friction to the user if it determines that a transaction is malicious? Users often click through things without reading, by allowing for a warning we can add friction at points that we really think a user should have a second glance.
```

SIPS/sip-11.md
```diff
@@ -0,0 +1,76 @@
...
+
+This warning could be injected at any point where there are insights provided.
+
+## Specification
+
+Please see [SIP-3](https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-3.md) for more information on the original transaction insights API.
+
+Please see the [@metamask/snaps-ui](https://github.com/MetaMask/snaps/blob/main/packages/snaps-ui/src/nodes.ts) package for more information on the `Component` type returned in the `OnTransactionResponse`.
```
Mrtenz, line 24: This should link to the UI SIP.

SIPS/sip-11.md
```diff
@@ -0,0 +1,76 @@
...
+The interface for the return value of an `onTransaction` export is:
+
+```typescript
+enum SeverityLevel {
+  Critical = 'critical',
+}
+
+interface OnTransactionResponse {
+  content: Component | null;
+  severity?: SeverityLevel;
```
Mrtenz, line 60: We should add a note that this SIP specifically just adds this field to the already existing API.

Mrtenz (dismissed):

SIPS/sip-11.md
```diff
@@ -0,0 +1,76 @@
+---
+sip: 11
+title: Transaction Insights V2 - Transaction Warnings
```
Mrtenz, line 3:
```suggestion
title: Providing severity for transaction insights
```
Or something like that?

SIPS/sip-11.md
hmalik88: I was doing that at first, but it's outdated the return type outlined there is not the most up-to-date one.

SIPS/sip-11.md
hmalik88: Yeah, i understand lol

SIPS/sip-11.md
```diff
@@ -0,0 +1,74 @@
...
+
+One of the biggest issues with wallet users is a loss of funds through executing some sketchy or seemingly innocuous transaction. With the benefit of the transaction insights API, a snap can already run analysis on an unsigned transaction payload. By adding `severity` to the existing API, we allow for the snap to provide extra friction to the user if it determines that a transaction is malicious. Users often click through things without reading, by allowing for a warning we can add friction at points that we really think a user should have a second glance.
+
+This warning could be injected at any point where there are insights provided.
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
```
Mrtenz, line 24: These key words are not used.

SIPS/sip-11.md
```diff
@@ -0,0 +1,68 @@
...
+```typescript
+import { OnTransactionHandler } from "@metamask/snap-types";
+
+export const onTransaction: OnTransactionHandler = async ({
+  transaction,
+  chainId,
+}) => {
+  const content = /* Get UI component with insights */;
+  const isContentCritical = /* Boolean checking if content is critical */
+  return isContentCritical ? { content, severity: 'critical' } : { content };
```
david0xd, line 41: Given that there is a type import at the top, can we avoid using `string` for a severity and use enum? Like the one defined in the example below?

SIPS/sip-11.md
```diff
@@ -0,0 +1,68 @@
...
+};
+```
+
+The interface for the return value of an `onTransaction` export is:
+
+```typescript
+enum SeverityLevel {
+  Critical = 'critical',
```
david0xd: Just thinking out loud... So, it is mentioned:
```
In future SIPs, the `SeverityLevel` enum can be expanded to include other levels that can be also be used to influence the UI in the extension.
```
I'm thinking if we can proactively define some other severity levels? It would be somehow a proof that this is required to be built in a dynamic extensive way which has multiple enum values, etc. Otherwise, this could be just a boolean `critical?: boolean`. I'm not suggesting that we use a boolean, I can sense already that this will expand, just maybe if we expand it now, it will be more clear about the direction in which this is going and allow the other use cases from the beginning.

For example, how I see it:
```typescript
enum SeverityLevel {
  Critical = 'critical',
  Danger = 'danger',
  Suspicious = 'suspicious',
}
```
Just an idea to think about.

Mrtenz: From a user's perspective there is no difference between a "suspicious" transaction and a "critical" transaction. In both cases the user (likely) shouldn't go through with the transaction. In the future we may want to handle these cases different in the UI, but for the initial implementation, it will be just critical or nothing.

david0xd (approved):

Montoya (approved):
LGTM

PR #111, opened 2023-10-19
Dynamic Permissions

FrederikBolding:
WIP

Closes https://github.com/MetaMask/snaps/issues/1820

### DIFF
--- a/SIPS/sip-14.md
+++ b/SIPS/sip-14.md
@@ -1,0 +1,70 @@
+---
+sip: 14
+title: Dynamic Permissions
+status: Draft
+discussions-to: 
+author: Frederik Bolding (@frederikbolding)
+created: 2023-10-19
+---
+
+## Abstract
+
+This SIP proposes changes to the snap manifest and new RPC methods that would allow snap developers to request additional permissions dynamically at runtime. This proposal outlines some of the details around this feature.
+
+## Motivation
+
+Snaps currently have to request all permissions that they plan to use at install-time. This becomes a problem when a snap wants to use many permissions as the installation experience suffers and the user has to either accept all permissions requested or deny the installation. This proposal provides an improvement to the experience by letting snaps request permissions at run-time as long as those permissions are statically defined in the manifest at build-time.
+
+## Specification
+
+> Formal specifications are written in Typescript. Usage of `CAIP-N` specifications, where `N` is a number, are references to [Chain Agnostic Improvement Proposals](https://github.com/ChainAgnostic/CAIPs).
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Manifest
+
+This SIP adds a new field to the snap manifest called `dynamicPermissions`.
+This field can be used in tandem with the existing `initialPermissions`, but keep in mind that permissions in this field are not granted by installation, they MUST be requested when needed.
+
+The new field can be specified as follows in a `snap.manifest.json` file:
+
+```json
+{
+  "initialPermissions": {
+    "endowment:transaction-insight": {}
+  },
+  "dynamicPermissions": {
+    "snap_dialog": {}
+  }
+}
+```
+
+### RPC Methods
+
+This SIP also proposes new RPC methods to manage these new permissions:
+
+#### snap_requestPermissions
+This RPC method SHOULD function as a subset of the existing `wallet_requestPermissions` RPC method and take the same parameters and have the same return value. This function MAY be a middleware that rewrites requests to `wallet_requestPermissions` if needed. 
+
+This RPC method MUST prompt the user to get consent for any requested permissions and MUST validate that the requested permissions are specified in the manifest before continuing its execution.
+
+#### snap_getPermissions
+This RPC method SHOULD be an alias for `wallet_getPermissions`, MAY be used by the snap for verifying whether it already has the permissions needed for operating. The return value and parameters SHOULD match the existing specification.
+
+#### snap_revokePermissions
+The RPC method parameters and return value are TBD.
+
+Note: This RPC method does not currently have a `wallet_` counterpart. Coordinate with dapp API team as they may be shipping one.
+
+This RPC method MUST validate that the permissions requested to be revoked does not contain or overlap with the `initialPermissions`.
+
+### Permission Caveats
+TBD
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE).


SIPS/sip-14.md
```diff
@@ -0,0 +1,81 @@
...
+title: Dynamic Permissions
+status: Draft
+discussions-to: 
+author: Frederik Bolding (@frederikbolding)
+created: 2023-10-19
+---
+
+## Abstract
+
+This SIP proposes changes to the snap manifest and new RPC methods that would allow snap developers to request additional permissions dynamically at runtime. This proposal outlines some of the details around this feature.
```
Mrtenz, line 12:
```suggestion
This SIP proposes changes to the Snap manifest and new RPC methods that would allow Snap developers to request additional permissions dynamically at runtime. This proposal outlines some of the details around this feature.
```

SIPS/sip-14.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Motivation
+
+Snaps currently have to request all permissions that they plan to use at install-time. This becomes a problem when a snap wants to use many permissions as the installation experience suffers and the user has to either accept all permissions requested or deny the installation. This proposal provides an improvement to the experience by letting snaps request permissions at run-time as long as those permissions are statically defined in the manifest at build-time.
```
Mrtenz, line 16:
```suggestion
Snaps currently have to request all permissions that they plan to use at install-time. This becomes a problem when a Snap wants to use many permissions as the installation experience suffers and the user has to either accept all permissions requested or deny the installation. This proposal provides an improvement to the experience by letting Snaps request permissions at run-time as long as those permissions are statically defined in the manifest at build-time.
```

SIPS/sip-14.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### snap_requestPermissions
+This RPC method SHOULD function as a subset of the existing `wallet_requestPermissions` RPC method and take the same parameters and have the same return value. This function MAY be a middleware that rewrites requests to `wallet_requestPermissions` if needed. 
+
+This RPC method MUST prompt the user to get consent for any requested permissions and MUST validate that the requested permissions are specified in the manifest before continuing its execution (including caveats matching).
+
+#### snap_getPermissions
+This RPC method SHOULD be an alias for `wallet_getPermissions`, MAY be used by the snap for verifying whether it already has the permissions needed for operating. The return value and parameters SHOULD match the existing specification.
+
+#### snap_revokePermissions
```
Mrtenz, line 72:
```suggestion
#### snap_revokePermissions

```

SIPS/sip-14.md
```diff
@@ -0,0 +1,81 @@
...
+The RPC method parameters and return value are TBD.
+
+Note: This RPC method does not currently have a `wallet_` counterpart. Coordinate with dapp API team as they may be shipping one.
```
Mrtenz, line 75: We should remove this line before merging.

Mrtenz (changes requested):

SIPS/sip-14.md
FrederikBolding: Yes, just adding until we talk to them

SIPS/sip-14.md
```diff
@@ -0,0 +1,86 @@
...
+    "endowment:transaction-insight": {}
+  },
+  "dynamicPermissions": {
+    "snap_dialog": {},
+    "snap_getBip44Entropy": [
+      {
+        "coinType": 1
+      },
+      {
+        "coinType": 3
```
Montoya, line 47: Does this mean that the Snap can never dynamically request a coinType that was not declared in the manifest? 

And what does this mean for caveats like `endowment:rpc`? Let's say dapp is `false` and snaps is `true.` Can the Snap dynamically request to make `dapp` true?

SIPS/sip-14.md
```diff
@@ -0,0 +1,86 @@
...
+created: 2023-10-19
+---
+
+## Abstract
+
+This SIP proposes changes to the Snap manifest and new RPC methods that would allow Snap developers to request additional permissions dynamically at runtime. This proposal outlines some of the details around this feature.
+
+## Motivation
+
+Snaps currently have to request all permissions that they plan to use at install-time. This becomes a problem when a Snap wants to use many permissions as the installation experience suffers and the user has to either accept all permissions requested or deny the installation. This proposal provides an improvement to the experience by letting Snaps request permissions at run-time as long as those permissions are statically defined in the manifest at build-time.
```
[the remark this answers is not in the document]
FrederikBolding: Yes, that is true. But `eth_accounts` is kind of an edge-case right now. We should probably require it being put into `dynamicPermissions`

SIPS/sip-14.md
FrederikBolding: As we have no way to merge caveats right now (and aren't planning to for v1 of this API) the limitations are:
- The snap can never request a caveat that is not predefined in the manifest
- The snap can also never request something that doesn't match exactly the definition in the manifest, you cannot even request coinType 1 and 3 separately.

Mrtenz: This should be mentioned in the SIP!

FrederikBolding: Is it not already?

SIPS/sip-14.md
```diff
@@ -0,0 +1,81 @@
...
+    ]
+  }
+}
+```
+
+### Permission caveats and merging
+
+In this initial version, duplicated permissions in `initialPermissions` and `dynamicPermissions` MUST NOT be allowed. A permission MUST only be able to exist in one of the manifest fields.
+
+Furthermore, permissions specified in `dynamicPermissions` MUST contain the caveats that will be requested at runtime and the permission request MUST fully match the caveats specified in the manifest.
```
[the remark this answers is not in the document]
FrederikBolding: Isn't that always implied?

SIPS/sip-14.md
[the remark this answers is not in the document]
Montoya: That would be a new permission? Right now we just require that they request `endowment:ethereum-provider`, so it's like a dynamic permission on top of a static one.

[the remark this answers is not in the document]
FrederikBolding: `eth_accounts` is an edge-case because it is technically a permission that is granted when using certain RPC methods with `endowment:ethereum-provider`. It is not inherently granted just by having the endowment.

We could choose to enforce that you need to have it as part of `dynamicPermissions` to grant it via `endowment:ethereum-provider`. We could also choose to keep it as the edge-case that it is.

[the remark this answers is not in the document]
Montoya: I do wonder if we have created a strange design pattern with the current approach. What if a Snap just wants `eth_accounts` and is requesting `endowment:ethereum-provider` just for that, and not using anything else that the endowment offers? Maybe it should be a separate endowment that must be defined in `dynamicPermissions` and is never in `initialPermissions`. Assuming this change would be in manifest version 0.2 and would not be backwards compatible.

Montoya (approved):
LGTM

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+title: Dynamic Permissions
+status: Draft
+discussions-to: https://github.com/MetaMask/SIPs/discussions/114
+author: Frederik Bolding (@frederikbolding)
+created: 2023-10-19
+---
+
+## Abstract
+
+This SIP proposes changes to the Snap manifest and new RPC methods that would allow Snap developers to request additional permissions dynamically at runtime. This proposal outlines some of the details around this feature.
```
Mrtenz, line 12:
```suggestion
This SIP proposes changes to the Snap manifest and new RPC methods that allows Snap developers to request additional permissions dynamically at runtime. This proposal outlines some of the details around this feature.
```

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Manifest
+
+This SIP adds a new field to the Snap manifest called `dynamicPermissions`.
+This field can be used in tandem with the existing `initialPermissions`, but keep in mind that permissions in this field are not granted by installation: They MUST be requested when needed. The field follows the same format as `initialPermissions`.
```
Mrtenz, line 31:
```suggestion
This field can be used in tandem with the existing `initialPermissions`, but permissions in this field are not granted by installation: They MUST be requested when needed. The field follows the same format as `initialPermissions`.
```

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+      },
+      {
+        "coinType": 3
+      }
+    ]
+  }
+}
+```
+
+### Permission caveats and merging
```
Mrtenz, line 54: Implementation detail?
```suggestion
### Permission caveats
```

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+
+In this initial version, duplicated permissions in `initialPermissions` and `dynamicPermissions` MUST NOT be allowed. A permission MUST only be able to exist in one of the manifest fields.
```
Mrtenz, line 56:
```suggestion
Duplicated permissions in `initialPermissions` and `dynamicPermissions` MUST NOT be allowed. A permission MUST only be able to exist in one of the manifest fields.
```

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+
+This MAY change in a future SIP.
+
+### RPC Methods
+
+This SIP proposes the following RPC methods to manage the dynamic permissions:
+
+#### snap_requestPermissions
+
+This RPC method SHOULD function as a subset of the existing `wallet_requestPermissions` RPC method and take the same parameters and have the same return value.
```
Mrtenz, line 68: Maybe link to the spec of `wallet_requestPermissions`.

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+
+This RPC method MUST prompt the user to get consent for any requested permissions and MUST validate that the requested permissions are specified in the manifest before continuing its execution (including caveats matching).
```
Mrtenz, line 70:
```suggestion
This RPC method MUST prompt the user to get consent for any requested permissions and MUST validate that the requested permissions are specified in the manifest before continuing its execution (including matching caveats).

```

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+
+#### snap_getPermissions
+
+This RPC method SHOULD be an alias for `wallet_getPermissions`, MAY be used by the snap for verifying whether it already has the permissions needed for operating. The return value and parameters SHOULD match the existing specification.
```
Mrtenz, line 74: Maybe link to the spec of `wallet_getPermissions`.

Mrtenz:
```suggestion
This RPC method SHOULD be an alias for `wallet_getPermissions`, and MAY be used by the Snap for verifying whether it already has the permissions needed for operating. The return value and parameters SHOULD match the existing specification.
```

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+
+#### snap_revokePermissions
+
+This RPC method SHOULD take a similar input to `wallet_requestPermissions`, an object keyed with permission names, where the values may contain caveats if applicable.
```
Mrtenz, line 78: Maybe link to the spec of `wallet_requestPermissions`.

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+            { "path": ["m", "0'", "0'"], "curve": "ed25519" }
+          ]
+        }
+      ]
+    }
+  }
+}
+```
+
+The caveat information passed SHOULD be ignored in the initial implementation of this. Instead of processing the caveats, the implementation SHOULD revoke the entire permission key. We will revisit this at a later time to make it more granular.
```
Mrtenz, line 101: Does it make sense to specify the caveats if it's ignored? Or is it for future compatibility?

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+
+This RPC method SHOULD return `null` if the permissions are revoked successfully and otherwise throw.
```
Mrtenz, line 103:
```suggestion
This RPC method SHOULD return `null` if the permissions are revoked successfully, or return an error otherwise.
```

SIPS/sip-14.md
```diff
@@ -0,0 +1,109 @@
...
+
+This RPC method MUST validate that the permissions requested to be revoked does not contain or overlap with the `initialPermissions`.
```
Mrtenz, line 105:
```suggestion
This RPC method MUST validate that the permissions to be revoked do not contain any permissions specified in `initialPermissions`. Only `dynamicPermissions` can be revoked.
```

SIPS/sip-14.md
FrederikBolding: The idea was that it would be future proof this way. We can make it more granular as we see fit

Mrtenz: I would probably just remove the caveat part. It just makes it more complicated for Snaps developers to use this method, doesn't it? In the future when we revisit this, Snaps developers can just change it to only revoke certain caveats, and we revoke everything if the object is empty or something.

Montoya (approved):
LGTM

PR #113, opened 2023-10-26
Snaps Pages

FrederikBolding:
Adds draft for the snaps home page feature.

Fixes https://github.com/MetaMask/MetaMask-planning/issues/1386

### DIFF
--- a/SIPS/sip-15.md
+++ b/SIPS/sip-15.md
@@ -19,7 +19,7 @@ This new surface differs from the existing custom UI surfaces as it isn't reacti
 
 ## Specification
 
-> Formal specifications are written in Typescript. Usage of `CAIP-N` specifications, where `N` is a number, are references to [Chain Agnostic Improvement Proposals](https://github.com/ChainAgnostic/CAIPs).
+> Formal specifications are written in Typescript.
 
 ### Language
 
--- a/SIPS/sip-15.md
+++ b/SIPS/sip-15.md
@@ -44,7 +44,7 @@ This permission is specified as follows in `snap.manifest.json` files:
 
 ### Snap Implementation
 
-When a user navigates to the snap home page, the `onHomePage` handler will be invoked. This handler MUST be used to generate the static UI content for the home page.
+When a user navigates to the snap home page, the `onHomePage` handler will be invoked. This handler MUST be used to generate the UI content for the home page.
 
 Any snap that wishes to expose a home page MUST implement the following API:
 
--- a/SIPS/sip-15.md
+++ b/SIPS/sip-15.md
@@ -1,0 +1,73 @@
+---
+sip: 15
+title: Snaps Home Page
+status: Draft
+discussions-to: 
+author: Frederik Bolding (@frederikbolding)
+created: 2023-10-26
+---
+
+## Abstract
+
+This SIP proposes a new API that allows snaps to surface a static UI to display data that may be useful to users of the snap. This proposal outlines some of the details around this feature.
+
+## Motivation
+
+Snaps currently tend to leverage separate websites for showing details relating to the status of the snap. For example, snaps that manage new blockchain accounts for users have to rely almost exclusively on these to show balances on new kinds of networks. This proposal aims to start improving on this by giving snaps a surface directly in the client where they can render custom UI. 
+
+This new surface differs from the existing custom UI surfaces as it isn't reactionary, instead of being triggered by an RPC call or a transaction being confirmed the user can choose to view the home page screen at any time.
+
+## Specification
+
+> Formal specifications are written in Typescript. Usage of `CAIP-N` specifications, where `N` is a number, are references to [Chain Agnostic Improvement Proposals](https://github.com/ChainAgnostic/CAIPs).
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Manifest
+
+This SIP specifies a permission named `endowment:snaps-home`.
+The permission signals to the platform that the snap wants to use the home page functionality. This MAY change routing in the client or change how the snap is displayed to the user.
+
+This permission is specified as follows in `snap.manifest.json` files:
+
+```json
+{
+  "initialPermissions": {
+    "endowment:snaps-home": {}
+  }
+}
+```
+
+### Snap Implementation
+
+When a user navigates to the snap home page, the `onHome` handler will be invoked. This handler MUST be used to generate the static UI content for the home page.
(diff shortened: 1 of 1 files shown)


SIPS/sip-15.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Motivation
+
+Snaps currently tend to leverage separate websites for showing details relating to the status of the snap. For example, snaps that manage new blockchain accounts for users have to rely almost exclusively on these to show balances on new kinds of networks. This proposal aims to start improving on this by giving snaps a surface directly in the client where they can render custom UI. 
+
+This new surface differs from the existing custom UI surfaces as it isn't reactionary, instead of being triggered by an RPC call or a transaction being confirmed the user can choose to view the home page screen at any time.
+
+## Specification
+
+> Formal specifications are written in Typescript. Usage of `CAIP-N` specifications, where `N` is a number, are references to [Chain Agnostic Improvement Proposals](https://github.com/ChainAgnostic/CAIPs).
```
Mrtenz, line 22:
```suggestion
> Formal specifications are written in Typescript.
```

SIPS/sip-15.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Manifest
+
+This SIP specifies a permission named `endowment:snaps-home`.
```
Mrtenz, line 32: I wonder if it makes sense to include "Snaps" in the permissions name? It's kind of implied. 😅

Mrtenz (changes requested):

SIPS/sip-15.md
hmalik88: I like endowment:home-page

FrederikBolding: Yeah, I agree.

SIPS/sip-15.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Abstract
+
+This SIP proposes a new API that allows snaps to surface a static UI to display data that may be useful to users of the snap. This proposal outlines some of the details around this feature.
+
+## Motivation
+
+Snaps currently tend to leverage separate websites for showing details relating to the status of the snap. For example, snaps that manage new blockchain accounts for users have to rely almost exclusively on these to show balances on new kinds of networks. This proposal aims to start improving on this by giving snaps a surface directly in the client where they can render custom UI. 
+
+This new surface differs from the existing custom UI surfaces as it isn't reactionary, instead of being triggered by an RPC call or a transaction being confirmed the user can choose to view the home page screen at any time.
```
Montoya, line 18: One could argue that it is reactionary, it is reacting to the user activating the page by opening it within the extension, or if the Snap deep links to this page from somewhere else. Every entry point is reactionary in some way. You could say it differs in that it is directly accessible by the user via the extension UI.

SIPS/sip-15.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Manifest
+
+This SIP specifies a permission named `endowment:home-page`.
```
Montoya, line 32: Let's say we eventually have different pages, would it be better to have: 

- `endowment:page-home`
- `endowment:page-tx-history`
- `endowment:page-send`

Instead of the pattern being : [location] - page?

SIPS/sip-15.md
```diff
@@ -0,0 +1,73 @@
...
+The permission signals to the platform that the snap wants to use the home page functionality. This MAY change routing in the client or change how the snap is displayed to the user.
```
Montoya, line 33: Alternatively `The permission signals to the platform that the snap wants to present a home page within the client UI.`

And is the second sentence needed? It's an implementation detail.

SIPS/sip-15.md
```diff
@@ -0,0 +1,73 @@
...
+{
+  "initialPermissions": {
+    "endowment:home-page": {}
+  }
+}
+```
+
+### Snap Implementation
+
+When a user navigates to the snap home page, the `onHome` handler will be invoked. This handler MUST be used to generate the static UI content for the home page.
```
Montoya, line 47: `onHome`? Not `onHomePage` or `onPageHome`?

SIPS/sip-15.md
FrederikBolding: It may not be

SIPS/sip-15.md
hmalik88: Maybe not reactionary to user input? Not sure if a view is considered input

SIPS/sip-15.md
FrederikBolding: I don't think it changes much, but we can make the change if you feel that it improves the pattern

SIPS/sip-15.md
FrederikBolding: I guess that is true, I was merely trying to explain how this is more surface-level than other snaps features that are "part of another flow"

SIPS/sip-15.md
hmalik88: I can see the above pages being useful for a protocol snap to integrate into our UI, but at that point we're permitting divergent ux within the extension and also contradicting our position of non-reactionary UI. I think having a home-page as a catch-all is a good intermediary step to direct integration to different parts of the extension i.e. tx-history, account balances etc.

SIPS/sip-15.md
hmalik88: Agreed

SIPS/sip-15.md
hmalik88: Agreeing with `onHomePage`

SIPS/sip-15.md
Montoya: These are bad examples but just assuming that there would eventually be different types of pages, maybe the pattern `page-[location]` is better than `[location]-page`

SIPS/sip-15.md
```diff
@@ -0,0 +1,73 @@
...
+{
+  "initialPermissions": {
+    "endowment:page-home": {}
+  }
+}
+```
+
+### Snap Implementation
+
+When a user navigates to the snap home page, the `onHomePage` handler will be invoked. This handler MUST be used to generate the static UI content for the home page.
```
Montoya, line 47: I would strike the word "static" here and then it's good to go

Montoya (changes requested):

hmalik88 (approved):

Montoya (approved):
LGTM

PR #120, opened 2023-11-15
Make Github account validation work as described in author spec

esterlus:
The spec defines an author as:

`Name Surname <email> (@github-username)`, e.g.: `John F. Kowalsky <john.kowalsky@domain.com> (@kowalsky)`

However the rule to determine the github user account only works with authors that do not have an email field.


FrederikBolding (approved):

Mrtenz (approved):
Thanks!

PR #122, opened 2023-11-19
SIP-18 Manage Networks Permissions

0xbeny:


hmalik88:
Thanks for this initial draft! I will take over this SIP since the spec part is missing and I have some deeper thoughts about this. I think we can expose a permission/method called `snap_manageEthereumChain` (that bypasses the approval request) that can be utilized with our lifecycle hooks to allow for some cool stuff like batch adding chains at various parts in the snaps lifecycle. I'll link the fleshed out SIP here.

PR #124, opened 2023-11-28
Added SIP that reproduces the checksum algorithm

### LINKED ISSUE #123: SIP for multi-file checksum algorithm
In the future we'd like to use [Decentralised Identifiers](https://www.w3.org/TR/did-core/) to refer to Snaps.

It's going to use multi-file checksum that we write to `snap.manfiest.json`, like so `did:snap:8uciDR11gmEm8QDs/sHYPeICmkpHOE4qVe+7zF0O7Ew=`.

The checksum algorithm is currently standardised only in [`snap-utils`](https://github.com/MetaMask/snaps/blob/main/packages/snaps-utils/src/checksum.ts) and is based on [`fast-json-stable-stringify`](https://github.com/MetaMask/snaps/blob/6e0257741c7eb0fb71df5826fcfabb7658abd03e/packages/snaps-utils/src/snaps.ts#L165) npm package.

We should write a SIP that provides all steps to make such a reproducible checksum

ritave:
This SIP describes an algorithm used to checksum a snap in a reproducible manner in a way that includes all files required to run the snap.

fixes #123


assets/sip-124/implementation/implementation.ts
```diff
@@ -0,0 +1,24 @@
+import stableStringify from "fast-json-stable-stringify";
+import { sha256 } from "@noble/hashes/sha256";
+import { concatBytes } from "@metamask/utils";
+import { base64 } from "@scure/base";
+
+export type VFile = { path: string; contents: string };
```
FrederikBolding, line 6:
```suggestion
export type VFile = { path: string; contents: string | Uint8Array };
```

assets/sip-124/implementation/implementation.ts
```diff
@@ -0,0 +1,24 @@
...
+  delete structure.result.source.shasum;
+  return { path, contents: stableStringify(structure) };
+}
+
+export function checksumFiles(manifest: VFile, auxiliary: VFile[]): string {
+  return base64.encode(
+    sha256(
+      concatBytes(
+        [stableManifest(manifest), ...auxiliary]
+          .sort((a, b) => (a.path < b.path ? -1 : 1))
```
FrederikBolding, line 19: We may wanna include the failure case for identical paths in this reference implementation and the SIP

SIPS/sip-124.md
```diff
@@ -0,0 +1,55 @@
+---
+sip: 124
```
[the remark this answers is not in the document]
ritave: There would already be a collision between open PR for sip-18 (#122) and this one.

I'd prefer to add a comment to SIP-1 about how the number of the SIP correlates to PR number

assets/sip-124/implementation/implementation.ts
ritave: Not used in the example, I'll leave it out to simplify the example

SIPS/sip-124.md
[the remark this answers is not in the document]
hmalik88: I think I like the incrementing scheme better, if you're following the growth of the platform you have a better mental model of where a SIP is in relation to others. One just needs to be mindful of the other open SIP PRs when marking the SIP number.

SIPS/sip-124.md
```diff
@@ -0,0 +1,55 @@
...
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Steps
+
+#### Checksum `snap.manifest.json`
+
+> This algorithm works on `snap.manifest.json` version `0.1`.
+
+<!-- TODO(ritave): Extract the algorithm from fast-json-stable-stringify and put it here -->
```
hmalik88, line 29: Feels weird to have a comment in the spec, can we either put the algorithm in here now or open an issue to add in later?

SIPS/sip-124.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. Parse `snap.manifest.json` into JavaScript object using `JSON.parse`.
+1. Delete `.result.source.shasum` field from `snap.manifest.json`.
```
hmalik88, line 32:
```suggestion
2. Delete `.result.source.shasum` field from `snap.manifest.json`.
```

SIPS/sip-124.md
```diff
@@ -0,0 +1,55 @@
...
+1. Convert rest of the structure back to JSON using [`fast-json-stable-stringify@^2.1.0`](https://www.npmjs.com/package/fast-json-stable-stringify) algorithm.
```
hmalik88, line 33:
```suggestion
3. Convert rest of the structure back to JSON using [`fast-json-stable-stringify@^2.1.0`](https://www.npmjs.com/package/fast-json-stable-stringify) algorithm.
```

SIPS/sip-124.md
```diff
@@ -0,0 +1,55 @@
...
+1. Checksum the resulting string using [auxiliary file](#checksum-auxiliary-files) algorithm.
```
hmalik88, line 34:
```suggestion
4. Checksum the resulting string using [auxiliary file](#checksum-auxiliary-files) algorithm.
```

PR #127, opened 2024-01-12
Add SIP-21: Snap-defined timeouts

FrederikBolding:
This PR adds SIP 21 called Snap-defined timeouts.

### DIFF
--- a/SIPS/sip-21.md
+++ b/SIPS/sip-21.md
@@ -13,7 +13,7 @@ This SIP describes a way for Snaps to customize the request timeouts that constr
 
 ## Motivation
 
-Snaps that have to do CPU intensive computation are currently limited by the default `1` minute timeout, causing certain use-cases to be unsupportable by Snaps (e.g. computing ZK proofs). By letting Snaps customize the request timeout (within reason) for each type of handler that they expose, the Snap developer gets more control over the user experience while the Snap platform lifecycle requirements can remain strict.
+Snaps that do CPU intensive computation are currently limited by the 1 minute timeout, causing certain use-cases to be unsupportable by Snaps (e.g., computing ZK proofs). By letting Snaps customize the request timeout (within reason) for each type of handler that they expose, the Snap developer gets more control over the user experience while the Snap platform lifecycle requirements can remain strict.
 
 ## Specification
 
--- a/SIPS/sip-21.md
+++ b/SIPS/sip-21.md
@@ -1,0 +1,52 @@
+---
+sip: 21
+title: Snap-defined timeouts
+status: Draft
+author: Frederik Bolding (@frederikbolding)
+created: 2024-01-12
+---
+
+## Abstract
+
+This SIP describes a way for Snaps to customize the request timeouts that constrain the Snap lifecycle. Effectively allowing the Snap to opt-in to slightly longer or shorter timeouts where needed.
+
+## Motivation
+
+Snaps that have to do CPU intensive computation are currently limited by the default `1` minute timeout, causing certain use-cases to be unsupportable by Snaps (e.g. computing ZK proofs). By letting Snaps customize the request timeout (within reason) for each type of handler that they expose, the Snap developer gets more control over the user experience while the Snap platform lifecycle requirements can remain strict.
+
+## Specification
+
+> Formal specifications are written in TypeScript.
+
+### Language
+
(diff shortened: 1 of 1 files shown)


SIPS/sip-21.md
```diff
@@ -0,0 +1,53 @@
...
+title: Snap-defined timeouts
+status: Draft
+discussions-to: https://github.com/MetaMask/SIPs/discussions/128
+author: Frederik Bolding (@frederikbolding)
+created: 2024-01-12
+---
+
+## Abstract
+
+This SIP describes a way for Snaps to customize the request timeouts that constrain the Snap lifecycle. Effectively allowing the Snap to opt-in to slightly longer or shorter timeouts where needed.
```
Mrtenz, line 12: What's the use case for a Snap requesting shorter timeouts?

SIPS/sip-21.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Motivation
+
+Snaps that have to do CPU intensive computation are currently limited by the default `1` minute timeout, causing certain use-cases to be unsupportable by Snaps (e.g. computing ZK proofs). By letting Snaps customize the request timeout (within reason) for each type of handler that they expose, the Snap developer gets more control over the user experience while the Snap platform lifecycle requirements can remain strict.
```
Mrtenz, line 16:
```suggestion
Snaps that do CPU intensive computation are currently limited by the 1 minute timeout, causing certain use-cases to be unsupportable by Snaps (e.g., computing ZK proofs). By letting Snaps customize the request timeout (within reason) for each type of handler that they expose, the Snap developer gets more control over the user experience while the Snap platform lifecycle requirements can remain strict.
```

SIPS/sip-21.md
```diff
@@ -0,0 +1,53 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Manifest
+
+This SIP specifies an addendum that MUST be applied to existing and future Snap handler permissions (`endowment:rpc`, `endowment:transaction-insight` etc.). The addendum is an optional caveat value called `maxRequestTime`.
+
+The value MUST be a valid integer between `5000` (5 seconds) and `180000` (3 minutes) and specifies the request timeout for the given handler in milliseconds. If no value is provided the default timeout of 1 minute (`30000` ms) MUST be used.
```
Mrtenz, line 32: Is 3 minutes enough for the ZK Snaps?

FrederikBolding: I think so, but we will get some feedback from snap devs before proceeding!

SIPS/sip-21.md
FrederikBolding: If a snap dev is confident that their snap runs faster than the 1 minute timeout and therefore confident that if it takes longer than 15 seconds something is wrong I think it makes sense for them to be able to use that as the timeout.

If something is actually wrong the user would be notified much quicker, making the UX slightly better.

SIPS/sip-21.md
Montoya: Yes my understanding is 3 minutes would be enough.

Montoya: From Rarimo: "3 minutes should be enough but we may want to do up to 5 minutes for more complex ZKPs in the future."

"Usually it takes about 40s on average hardware and up to 2 minutes on slow devices to generate a ZKP at https://robotornot.rarimo.com/" 

"We’ll have estimates for a bit more complex ZKPs in our new project in the next two weeks"

SIPS/sip-21.md
```diff
@@ -0,0 +1,53 @@
...
```
ihordiachenko, line 32:
```suggestion
The value MUST be a valid integer between `5000` (5 seconds) and `180000` (3 minutes) and specifies the request timeout for the given handler in milliseconds. If no value is provided the default timeout of 1 minute (`60000` ms) MUST be used.
```
Fixed typo

FrederikBolding: Thanks!

Mrtenz (approved):

hmalik88 (approved):

PR #134, opened 2024-03-21
Create SIP-22: Deep Links

Montoya:
I have attempted to describe a URL schema for deep links as well as a method for using them.

### DIFF
--- a/SIPS/sip-22.md
+++ b/SIPS/sip-22.md
@@ -37,7 +37,9 @@ Where:
   - For the `snap` authority, the path starts with the snap ID and has the following paths available to it:
     - `/home` - leads to the snap's [home page](/SIPS/sip-15.md) (which is its settings page if it doesn't have a home page)
 
-**Note:** In the future, fragments can potentially be used for navigation to specific portions of a page and params can be provided from a calling method for exports such as `onHome`.
+**Note:** 
+1. In the future, fragments can potentially be used for navigation to specific portions of a page and params can be provided from a calling method for exports such as `onHome`.
+2. This URL scheme assumes parity between `client` locations.
 
 ### Examples
 
--- a/SIPS/sip-22.md
+++ b/SIPS/sip-22.md
@@ -31,7 +31,7 @@ A URL for a deep link uses the format:
 Where: 
 
 - `[Snap ID]` refers to the canonical ID of the installed Snap
-- `page` refers to a page inside of the application 
+- `page` refers to a [Snap page](/SIPS/sip-15.md)
 - `[page name]` refers to the canonical name of an individual page 
 
 ### Method for Navigating to a Deep Link
--- a/SIPS/sip-22.md
+++ b/SIPS/sip-22.md
@@ -8,11 +8,11 @@ created: 2024-03-20
 
 ## Abstract
 
-This SIP describes a URL scheme for Snaps entry points. This scheme can be used by Snap or dapp methods to trigger the entry point and navigate the user to that location. 
+This SIP describes a URL scheme for Snaps entry points. This scheme can be used by Snap methods to navigate the user to that location in-client and trigger the entry point. 
 
 ## Motivation
 
-While some Snap entry points are reactive, meaning they can be triggered by a method, others can only be triggered by a user navigating to the location where that entry point is displayed. In order for a Snap or dapp to programmatically navigate a user to that entry point, the Snap or dapp needs a way to identify that location. The location is identified by a URL scheme, which is called a _deep link_.
+While some Snap entry points are reactive, meaning they can be triggered by a method, others can only be triggered by a user navigating to the location where that entry point is displayed. In order for a Snap to programmatically navigate a user to that entry point, the Snap needs a way to identify that location. The location is identified by a URL scheme, and an in-client link pointing to this location is called a _deep link_.
 
 ## Specification
 
@@ -26,18 +26,17 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
 
 A URL for a deep link uses the format: 
 
-`metamask://[Snap ID]/page/[page name]`
+`[Snap ID]/page/[page name]`
 
 Where: 
 
-- `metamask` refers to the host application 
 - `[Snap ID]` refers to the canonical ID of the installed Snap
 - `page` refers to a page inside of the application 
 - `[page name]` refers to the canonical name of an individual page 
 
 ### Method for Navigating to a Deep Link
 
-In order for a Snap or dapp to navigate a user to a deep link, a method is provided: 
+In order for a Snap to navigate a user to a deep link, a method is provided: 
 `snap_open`. 
 This method takes one parameter, which is the URL of the deep link: 
 
@@ -45,7 +44,7 @@ This method takes one parameter, which is the URL of the deep link:
 await snap.request({
   method: 'snap_open',
   params: {
-    url: 'metamask://npm:@consensys/starknet-snap/page/home'
+    url: 'npm:@consensys/starknet-snap/page/home'
   },
 });
 ```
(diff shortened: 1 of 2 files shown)


SIPS/sip-22.md
```diff
@@ -0,0 +1,119 @@
...
+
+- `metamask` refers to the host application 
+- `[Snap ID]` refers to the canonical ID of the installed Snap
+- `page` refers to a page inside of the application 
+- `[page name]` refers to the canonical name of an individual page 
+
+### Method for Navigating to a Deep Link
+
+In order for a Snap or dapp to navigate a user to a deep link, a method is provided: 
+`snap_open`.
```
Mrtenz, line 41: `snap_*` methods are only supported by Snaps. We may want to have a `wallet_*` method instead?

SIPS/sip-22.md
```diff
@@ -0,0 +1,119 @@
...
+
+In order for a Snap or dapp to navigate a user to a deep link, a method is provided: 
+`snap_open`. 
+This method takes one parameter, which is the URL of the deep link: 
+
+```typescript
+await snap.request({
+  method: 'snap_open',
+  params: {
+    url: 'metamask://npm:@consensys/starknet-snap/page/home'
```
Mrtenz, line 48: This is not a valid URL. It will navigate to `metamask://consensys/starknet-snap/page/home`.

SIPS/sip-22.md
```diff
@@ -0,0 +1,119 @@
...
+A URL for a deep link uses the format: 
+
+`metamask://[Snap ID]/page/[page name]`
+
+Where: 
+
+- `metamask` refers to the host application 
+- `[Snap ID]` refers to the canonical ID of the installed Snap
+- `page` refers to a page inside of the application 
+- `[page name]` refers to the canonical name of an individual page
```
Mrtenz, line 36: Maybe add a list of pages that can be used.

SIPS/sip-22.md
```diff
@@ -0,0 +1,119 @@
...
+
+A URL for a deep link uses the format: 
+
+`metamask://[Snap ID]/page/[page name]`
+
+Where: 
+
+- `metamask` refers to the host application 
+- `[Snap ID]` refers to the canonical ID of the installed Snap
+- `page` refers to a page inside of the application
```
hmalik88, line 35: Seems unnecessary given that the deep link is always inside the application.

```suggestion
```

SIPS/sip-22.md
Montoya: Actually this makes me think the Snap should have the method internally and expose a custom RPC method to dapps if it wants to allow dapps to do this.

SIPS/sip-22.md
Montoya: Oh well, maybe we have to encode the Snap ID then. I was hoping not to do that but seems like we don't have much choice....

Montoya:
I think I have a different idea about how to do this now. Will make a different proposal.

SIPS/sip-22.md
```diff
@@ -0,0 +1,119 @@
...
+    url: 'metamask://npm:@consensys/starknet-snap/page/home'
+  },
+});
+```
+
+### Examples
+
+The proposed URL for the Starknet Snap home page: 
+
+`metamask://npm:@consensys/starknet-snap/page/home`
```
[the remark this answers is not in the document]
Montoya: Tough question. One possibility is allowing URLs like: 

`localhost:8000/page/home` 

Another is to say that any URL without the snap ID, like just `/page/home` will be interpreted as referring to the current Snap, and will work on both localhost and production. 

I like either approach, what do others think?

Montoya:
OK I have updated this based on what I think is the right approach. The only thing that is missing is that ideally the request to open a page could also pass along some parameters. For example, let's say a Snap has multiple features in its home page, like 1) swap 2) bridge 3) stake. If a Snap wants to open its home page and navigate to one of those interfaces, it will need to be able to pass a parameter that can be processed by the homepage. The request could look like this: 

```typescript
await snap.request({
  method: 'snap_open',
  params: {
    url: 'npm:@consensys/starknet-snap/page/home',
    params: {...}
  },
});
```

Would that be a good approach?

SIPS/sip-22.md
Montoya: I made the method specific to Snaps but I am not actually certain whether it would be best to make it a `snap_*` or `wallet_*` method. Would like to hear your perspective based on the latest draft.

SIPS/sip-22.md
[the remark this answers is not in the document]
hmalik88: What do you mean by current snap?

SIPS/sip-22.md
hmalik88: My thoughts regarding this is if we want to expose this functionality to dapps, then we should make this a `wallet_*` method. Reason is because otherwise you end up having an extraneous request to the snap. It now has to be called for the invoke snap and then it's called for the `onHomePage` call.

hmalik88:
> OK I have updated this based on what I think is the right approach. The only thing that is missing is that ideally the request to open a page could also pass along some parameters. For example, let's say a Snap has multiple features in its home page, like 1) swap 2) bridge 3) stake. If a Snap wants to open its home page and navigate to one of those interfaces, it will need to be able to pass a parameter that can be processed by the homepage. The request could look like this:
> 
> ```ts
> await snap.request({
>   method: 'snap_open',
>   params: {
>     url: 'npm:@consensys/starknet-snap/page/home',
>     params: {...}
>   },
> });
> ```
> 
> Would that be a good approach?

You can have the params in the url, condense the request a little.

It makes sense to have params for the urls, we might have to update the SIP that defines pages to accommodate params.

SIPS/sip-22.md
hmalik88: Might have to encode the snap ID, as I said in standup the url is still not good without `metamask://`, maybe we can ship a url builder function to make it easier for devs.

hmalik88:
On second thought you can just use anchors if all you're looking for out of the params is to navigate to a certain part of a page

hmalik88:
Changed the URL scheme based on some more convos and reduced the scope of the SIP to just define the scheme, in a later SIP (or MIP?) we can spec the RPC navigation method.

SIPS/sip-22.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Examples
+
+The proposed URL for the Starknet Snap home page: 
+
+`metamask://snap/npm:@consensys/starknet-snap/home`
+
+The URL for navigating to the extension's home page:
+
+`metamask://extension/`
```
Montoya, line 50: This is good but I don't want to make it client specific. Why not: 

`metamask://client/`

That would work on mobile too.

Montoya: This would also align with the terminology we use elsewhere like `snap_getClientStatus`

hmalik88: Oh right, forgot we have something called mobile 😂

SIPS/sip-22.md
```diff
@@ -0,0 +1,60 @@
...
+- `[authority]` refers to either `client` or `snap` (in the current case, `client` can mean either the extension or mobile version of MetaMask)
+- `[path]` refers to the entire path which depending on the authority can be different
+  - For the `client` authority, the following paths are available:
+    - `/` - links to the client's home page
+  - For the `snap` authority, the path starts with the snap ID and has the following paths available to it:
+    - `/home` - leads to the snap's [home page](/SIPS/sip-15.md) (which is its settings page if it doesn't have a home page)
+
+**Note:** 
+1. In the future, fragments can potentially be used for navigation to specific portions of a page and params can be provided from a calling method for exports such as `onHome`.
+2. This URL scheme assumes parity between `client` locations.
```
Mrtenz, line 42: Nit:
```suggestion
> [!NOTE]
> 1. In the future, fragments can potentially be used for navigation to specific portions of a page and params can be provided from a calling method for exports such as `onHome`.
> 2. This URL scheme assumes parity between `client` locations.
```

SIPS/sip-22.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Examples
+
+The proposed URL for the Starknet Snap home page: 
+
+`metamask://snap/npm:@consensys/starknet-snap/home`
+
+The URL for navigating to the extension's home page:
```
Mrtenz, line 50: Nit:
```suggestion
The URL for navigating to the client's home page:
```

SIPS/sip-22.md
```diff
@@ -0,0 +1,60 @@
...
+
+`metamask://client/`
+
+### Using Deep Links
+
+A Snap can currently use navigation from within the snaps ui `Link` component. In the future, a SIP will be drafted to allow for programmatic navigation with another RPC method.
```
Mrtenz, line 56: We don't need to speculate about future SIPs here imo.
```suggestion
A Snap will be able to navigate to these links using the SIP-7 `Link` component.
```

SIPS/sip-22.md
hmalik88: The language was intentional because it is actually an example url of how you would navigate to the extension's home page.

SIPS/sip-22.md
```diff
@@ -0,0 +1,60 @@
+---
+sip: 22
+title: MetaMask URL scheme
+status: Draft
+author: Christian Montoya (@Montoya), Hassan Malik(@hmalik88)
```
Mrtenz, line 5: Nit:
```suggestion
author: Christian Montoya (@Montoya), Hassan Malik (@hmalik88)
```

SIPS/sip-22.md
Mrtenz: Not on mobile where this will eventually be supported.

GuillaumeRx (approved):

PR #147, opened 2024-08-30
SIP-26: Non-EVM protocol support

danroc:
## Abstract

This SIP presents an architecture to enable Snaps to expose blockchain-specific methods to dapps, extending MetaMask's functionality to support a multichain ecosystem.

## Motivation

Currently, MetaMask is limited to EVM-compatible networks. This proposal aims to empower developers, both first- and third-party, to use Snaps to add native support for non-EVM-compatible chains within MetaMask.

Fixes: https://github.com/MetaMask/accounts-planning/issues/542


SIPS/sip-26.md
```diff
@@ -0,0 +1,168 @@
...
+ */
+function resolveAccountAddress(request: MultichainRequest): string | undefined;
+```
+
+There must be only one Account Resolution Snap registered per chain to prevent
+ambiguity in the account resolution process.
+
+To identify which Account Resolution Snap should be used for a given request,
+Account Address Resolution Snaps MUST list the supported chains in their
+manifest file:
```
gantunesr, line 115: Nit,

```suggestion
To identify which Account Resolution Snap should be used for a given request,
Account Address Resolution Snaps MUST list the supported chains in their
manifest file using the [CAIP-2][caip-2] format for chain IDs:
```

SIPS/sip-26.md
```diff
@@ -0,0 +1,168 @@
...
+  for signing requests and managing accounts.
+
+- **Protocol Snaps**: Snaps that implement protocol methods that do not require
+  an account to be executed.
+
+- **RPC Router**: Native component that forwards non-signing requests to the
+  appropriate Protocol Snap or native implementation.
+
+- **Account Router**: Native component that forwards signing requests to the
+  appropriate Account Snap or native implementation.
```
gantunesr, line 50: I see that the `Account Router` is the one that registers signing methods to the `RPC Router`, but how are the signing methods registered into the `Account Router`?

gantunesr (commented):
Left a couple of small comments, overall this looks like a good improvement to handle multichain requests and scale the wallet beyond EVM chains

SIPS/sip-26.md
danroc: They are listed in the account object in the `methods` field. 

But it made me think of another problem: the account doesn't list the supported chains (there is a method to get this information, but we should probably move it into the account object).

SIPS/sip-26.md
```diff
@@ -0,0 +1,168 @@
...
+
+- **Account Address Resolution Snaps**: Snaps that implement the
+  `resolveAccountAddress` method to extract the account address from the
+  request object.
```
rekmarks, line 54: Can these Snaps be the same as Account and Protocol Snaps, or do they have to be separate?

SIPS/sip-26.md
```diff
@@ -0,0 +1,168 @@
...
+The diagram below represents a high-level architecture of how the Account
+Router integrates with the RPC Router, Account Snaps, and Protocol Snaps.
+
+![High-level architecture](../assets/sip-26/components-diagram.png)
+
+- **Account Snaps**: Snaps that implement the Keyring API and are responsible
+  for signing requests and managing accounts.
+
+- **Protocol Snaps**: Snaps that implement protocol methods that do not require
+  an account to be executed.
```
adonesky1, line 44: I thought per our discussions with @rekmarks we'd determined that we can't enforce separation of Account + Protocol Snaps?
Accounts SIP (1)

SIPS/sip-26.md
```diff
@@ -0,0 +1,168 @@
...
+execute and MUST list their supported methods in their manifest file:
+
+```json5
+"initialPermissions": {
+  "endowment:protocol-methods": {
+    "chains": {
+      "<chain_id_1>": [
+        // List of supported methods
+      ]
+    }
```
shanejonas, line 147: this should probably resemble caip-25 and the permission we just created and migrated the old account permissions to:

```

   "eip155:1": {
      "methods": ["eth_sendTransaction", "eth_signTransaction", "eth_sign", "get_balance", "personal_sign"],
      "notifications": []
   },
```

SIPS/sip-26.md
```diff
@@ -0,0 +1,168 @@
...
+
+```json5
+"initialPermissions": {
+  "endowment:account-address-resolver": {
+    "chains": [
+      "<chain_id_1>",
+      "<chain_id_2>",
+      // ...
+    ]
+  }
```
shanejonas, line 125: same as below, should resemble caip-25 permission: https://github.com/MetaMask/SIPs/pull/147/files#r1777172390

SIPS/sip-26.md
```diff
@@ -0,0 +1,168 @@
...
+// This will notify the Account Router that a new account was created, and the
+// Account Router will register this account as available for signing requests
+// using the `eth_signTypedData_v4` method.
+await snap.request({
+  method: 'snap_manageAccounts',
+  params: {
+    method: 'notify:accountCreated',
+    params: {
+      account: {
+          id: '74bb3393-f267-48ee-855a-2ba575291ab0',
```
shanejonas, line 74: the snap itself creates the id? we probably shouldn't leave these `id`s up to the developer.

SIPS/sip-26.md
adonesky1: Oh I see the language saying these can be combined. I think its more confusing to represent them as (sortof) separate in the diagram. 

As I believe I said in our meeting I don't think we're really getting anything by having these non signing protocol snaps that implement non signing methods if the signing snaps can trivially handle those requests too.

In that case I feel like its clearer to call them something more general than "Account Snaps" since they are not just handling accounts. Maybe "Protocol Snaps" or "CAIP-2 Snaps" (though I'd prefer something that clearly states its purpose without external reference).

SIPS/sip-26.md
```diff
@@ -0,0 +1,168 @@
...
+          type: 'eip155:eoa',
+          address: '0x1234567890123456789012345678901234567890',
```
shanejonas, line 76: this should be a caip-10 address to include the chainid which is missing context here when creating an account:

```
eip155:1:0xab16a96D359eC26a11e2C2b3d8f8B8942d5Bfcdb
```

SIPS/sip-26.md
```diff
@@ -0,0 +1,168 @@
...
+          methods: ['eth_signTypedData_v4'],
+          options: {},
```
shanejonas, line 78: what would go in `options`?

SIPS/sip-26.md
adonesky1: Yea agreed. We will need some way to associate this account with the appropriate [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-2.md) identifier for routing purposes, whether its making these CAIP-10 addresses or adding a CAIP-2 property on the object.

SIPS/sip-26.md
rekmarks: To restate what I said in a meeting: As I read it, they are just two separate permissions that a Snap may or may not put together, and some Snaps (e.g. hardware wallet or SCAs) may want to only ship the account part.

SIPS/sip-26.md
danroc: This is part of the Keyring API and is used by Snap devs to store implementation-dependent information about the account. An example would be to use it to store the tresholds, index, and other metadata used by MPC accounts.

SIPS/sip-26.md
danroc: Yeah, they are UUIDv4, the extension enforces that they are unique when an account is registered.

SIPS/sip-26.md
```diff
@@ -0,0 +1,163 @@
...
+![High-level architecture](../assets/sip-26/components-diagram.png)
+
+- **Account Snaps**: Snaps that implement the Keyring API and are responsible
+  for signing requests and managing accounts.
+
+- **Protocol Snaps**: Snaps that implement protocol methods that do not require
+  an account to be executed.
+
+- **RPC Router**: Native component that forwards RPC requests to the
+  appropriate Protocol Snap, Account Snap or native implementation.
```
adonesky1, line 47: I believe its more likely that requests that can be handled by native implementation will not reach the RPC Router
```suggestion
  appropriate Protocol Snap or Account Snap.
```

adonesky1: This also appears to match @FrederikBolding's [current implementation](https://github.com/MetaMask/snaps/pull/2875/files#diff-5697fc531248874f2ce02e26e9e13fee8222753070a8c8e107de6c281ba220afR246)

SIPS/sip-26.md
adonesky1: 🤔 Do we have a plan for how to handle types that are not `eip155:eoa` then? Seems like we should include a CAIP-2 scope(s) property here?

adonesky1: is there documentation we could link to where this `type` property is defined/scoped?

SIPS/sip-26.md
```diff
@@ -0,0 +1,163 @@
...
+    },
+  },
+});
+```
+
+Similar events are available to notify about the removal and update of
+accounts: `notify:accountRemoved` and `notify:accountUpdated`.
+
+Additionally, the Snaps Keyring expects the Account Snap to implement the
+Keyring API so it can forward signing requests to it through the
```
adonesky1, line 111: can we link to this `Keyring API` here?

adonesky1: Or is `Keyring API` just describing the `snaps_manageAccounts` method?

SIPS/sip-26.md
```diff
@@ -0,0 +1,163 @@
...
+### High-Level architecture
+
+The diagram below represents a high-level architecture of how the RPC Router
+integrates inside MetaMask to allow Snaps to expose protocol methods to dapps
+and the MetaMask clients.
+
+![High-level architecture](../assets/sip-26/components-diagram.png)
+
+- **Account Snaps**: Snaps that implement the Keyring API and are responsible
+  for signing requests and managing accounts.
```
adonesky1, line 41:
```suggestion
- **Account Snaps**: Snaps that implement the [Keyring API][keyring-api] and are responsible
  for signing requests and managing accounts.
```

FrederikBolding: Resolved this in my PR

SIPS/sip-26.md
FrederikBolding: Resolved this in my PR

FrederikBolding (approved):

Mrtenz (approved):

SIPS/sip-26.md
```diff
@@ -0,0 +1,194 @@
...
+interface OnProtocolRequestArguments {
+  origin: string;
+  scope: CaipChainId;
+  request: JsonRpcRequest;
+}
+```
+
+`origin` - The origin making the protocol request (i.e. a dapp).
+
+`scope` - The scope of the request, currently a chain ID as defined by the [CAIP-2 specification][caip-2].
```
adonesky1, line 181: nit
```suggestion
`scope` - The [CAIP-2][caip-2] chainId the request is targeting.
```

SIPS/sip-26.md
```diff
@@ -0,0 +1,194 @@
...
+
+### Components
+
+Here is a brief description of the components involved in this architecture
+which will require to be implemented or modified.
+
+#### RPC Router
+
+The RPC Router will be a new native component responsible for routing JSON-RPC
+requests to the appropriate Snap or keyring.
```
adonesky1, line 66: Is it confusing to separate `keyring` from Snap here since all the keyrings we route to are Snap keyrings?
```suggestion
requests to the appropriate Snap.
```

SIPS/sip-26.md
```diff
@@ -0,0 +1,194 @@
...
+
+To route a request, the RPC Router MUST extract the method name and chain ID
```
adonesky1, line 68: [nit]
```suggestion
To route a request, the RPC Router MUST extract the method name and [CAIP-2 chainId][caip-2]
```

SIPS/sip-26.md
```diff
@@ -0,0 +1,194 @@
...
+  };
+};
+```
+
+The implementation MUST return a value of the type `{ address: string }` or `null`.
+
+#### Protocol Snaps
+
+Protocol Snaps implement and expose methods that do not require an account to
+execute and MUST list their supported methods in their manifest file:
```
adonesky1, line 136:
```suggestion
execute and MUST list their supported methods and notifications in their manifest file:
```

SIPS/sip-26.md
```diff
@@ -0,0 +1,194 @@
...
+await snap.request({
+  method: "snap_manageAccounts",
+  params: {
+    method: "notify:accountCreated",
+    params: {
+      account: {
+        id: "74bb3393-f267-48ee-855a-2ba575291ab0",
+        type: "eip155:eoa",
+        address: "0x1234567890123456789012345678901234567890",
+        methods: ["eth_signTypedData_v4"],
```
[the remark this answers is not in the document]
adonesky1: Also I'm still wondering how we plan to standardize non-evm account types. Currently our [documentation](https://docs.metamask.io/snaps/reference/keyring-api/account-management/objects/#keyringaccount) only appears to describe `eip155:eoa` or `eip155:erc4337`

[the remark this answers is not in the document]
FrederikBolding: I have been requesting this as well, we need a more generic type that allows any protocol. IIRC @danroc and team are thinking about how to do that.

SIPS/sip-26.md
```diff
@@ -0,0 +1,196 @@
...
+  method: "keyring_resolveAccountAddress";
+  params: {
+    scope: CaipChainId;
+    request: JsonRpcRequest;
+  };
+};
+```
+`scope` - The [CAIP-2][caip-2] chainId the request is targeting
+
+`request` - A `JsonRpcRequest` containing strictly JSON-serializable values.
```
FrederikBolding, line 132:
```suggestion
`request` - A `JsonRpcRequest` containing strictly JSON-serializable values.

```

FrederikBolding (approved):

adonesky1 (approved):

PR #148, opened 2024-09-18
SIP-27: Accounts Metadata Request

ritave:
This SIP proposes extension to the [Keyring API](https://github.com/MetaMask/accounts/blob/main/packages/keyring-api), with a method called `keyring_listAccountsAll` that exposes all accounts usable by the user in the extension.

It's result is based on [`keyring_listAccounts`](https://github.com/MetaMask/accounts/blob/955192def55db49c19a0334eccacfa52a7071e34/packages/keyring-api/src/api/account.ts#L38)

### DIFF
--- a/SIPS/sip-27.md
+++ b/SIPS/sip-27.md
@@ -1,0 +1,68 @@
+---
+sip: 27
+title: Accounts Metadata Request
+status: Draft
+author: Olaf Tomalka (@ritave)
+created: 2024-09-17
+---
+
+## Abstract
+
+This SIP allows snaps to retrieve metadata related to accounts that exist in the extension.
+
+## Motivation
+
+The intention of this SIP is to allow snaps providing new accounts for new chains to be able to list all accounts when selecting one during transfers.
+
+## Specification
+
+> Indented sections like this are considered non-normative.
+
+> Usage of `CAIP-N` specifications, where `N` is a number, are references to [Chain Agnostic Improvement Proposals](https://github.com/ChainAgnostic/CAIPs).
+
+Formal specification of the proposed changes in the SIP. The specification must be complete enough that an implementation can be created from it.
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Manifest
+
+This SIP specifies a permission named `keyring_listAccountsAll`. This permission grants a snap the ability to retrieve account metadata through an RPC call.
+
+The permission is specified in `snap.manifest.json` as follows:
+
+```json
+{
+  "initialPermissions": {
+    "keyring_listAccountsAll": {}
+  }
+}
+```
+
+### RPC Method
+
+> Notice
+
+This SIP exposes a new RPC method called `keyring_listAccountsAll` with no additional parameters.
+
+The RPC call returns with the following data:
+
+```typescript
+type Account = {
+  id: string; // An extension-specific unique ID
+  address: string; // A blockchain specific public address for the account.
+  name: string; // User-given nickname for the account in the extension
+  chains: string[]; //
+};
+
+type Keyring_ListAccountsAllResult = Account[];
+```
+
+> Notice that multiple `Account`s can have the same `address`, for example when there are two hardware wallets using the same seed.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE).


SIPS/sip-27.md
```diff
@@ -0,0 +1,64 @@
...
+sip: 27
+title: Accounts Metadata Request
+status: Draft
+author: Olaf Tomalka (@ritave)
+created: 2024-09-17
+---
+
+## Abstract
+
+This SIP allows snaps to retrieve metadata related to accounts that exist in the extension.
```
Mrtenz, line 11:
```suggestion
This SIP allows Snaps to retrieve metadata related to accounts that exist in the client.
```

SIPS/sip-27.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Motivation
+
+The intention of this SIP is to allow snaps providing new accounts for new chains to be able to list all accounts when selecting one during transfers.
```
Mrtenz, line 15:
```suggestion
The intention of this SIP is to allow Snaps providing new accounts for new chains to be able to list all accounts when selecting one during transfers.
```

Can you add _why_ this is necessary as well?

SIPS/sip-27.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Manifest
+
+This SIP specifies a permission named `keyring_listAccountsAll`. This permission grants a snap the ability to retrieve account metadata through an RPC call.
```
Mrtenz, line 31: What do you think of this?
```suggestion
This SIP specifies a permission named `keyring_listAllAccounts`. This permission grants a snap the ability to retrieve account metadata through an RPC call.
```

Mrtenz, line 31:
```suggestion
This SIP specifies a permission named `keyring_listAccountsAll`. This permission grants a Snap the ability to retrieve account metadata through an RPC call.
```

SIPS/sip-27.md
```diff
@@ -0,0 +1,64 @@
...
+### RPC Method
+
+This SIP exposes a new RPC method called `keyring_listAccountsAll` with no additional parameters.
+
+The RPC call returns with the following data:
+
+```typescript
+type KeyringAccount = {
+  id: string; // An extension-specific unique ID
+  address: string; // A blockchain specific public address for the account.
```
Mrtenz, line 52: Should this be a CAIP-10 address?

Mrtenz (changes requested):
It's a bit unclear from this SIP whether this applies to all accounts in MetaMask (including Ethereum accounts not created through an account Snap), or just account Snap accounts. Maybe you can clarify that too.

SIPS/sip-27.md
ritave: This address can exist on multiple chains at the same time. Keyrings initially had a `chains` property, but it's not implemented

SIPS/sip-27.md
ritave: I specifically don't want it to be mistaken with existing `keyring_listAccounts`, which only lists accounts managed by the Snap. IMHO, by appending it at the end it's more visible,

Mrtenz: I wonder if it makes sense to prefix this with `keyring_` in the first place. So far these APIs are only used by account Snaps.

SIPS/sip-27.md
Mrtenz: Then how would you identify which network the address is tied to? 🤔

ritave: The only data available in MetaMask currently is the `type` property.

SIPS/sip-27.md
ritave: If we want to display _all_ accounts, we have to support how keyring manages and stores them in the extension, so this API mimics keyring internals almost 1:1.

SIPS/sip-27.md
```diff
@@ -0,0 +1,70 @@
...
+  name: string; // User-given nickname for the account in the extension
+  type: string; // Blockchain specific type of the account. For example "eip155:erc4337"
+};
+
+type Keyring_ListAccountsAllResult = KeyringAccount[];
+```
+
+> Notice that multiple `Account`s can have the same `address`, for example when there are two hardware wallets using the same seed.
+
+The call returns data concerning all accounts available in the extension, both built-in Ethereum accounts as well as accounts managed by this, and other Snaps.
```
Mrtenz, line 66:
```suggestion
The call returns data concerning all accounts available in the client, both built-in Ethereum accounts as well as accounts managed by this, and other Snaps.
```

SIPS/sip-27.md
```diff
@@ -0,0 +1,70 @@
...
+
+### RPC Method
+
+This SIP exposes a new RPC method called `keyring_listAccountsAll` with no additional parameters.
+
+The RPC call returns with the following data:
+
+```typescript
+type KeyringAccount = {
+  id: string; // An extension-specific unique ID
```
Mrtenz, line 55: Does it make sense to expose this? What do you envision we use it for?

SIPS/sip-27.md
Mrtenz: I guess my question is, should we combine `type` and `address` into one field with a CAIP-10 address.

SIPS/sip-27.md
ritave: When you have two accounts with the same address (eg 2 different hardware wallets with the same seed), you need to differentiate them somehow.

SIPS/sip-27.md
ritave: `type` exposes more information than just the network the address is on.

Accounts team envisioned the following types:
```typescript
type AccountType =
	| 'eip155:eoa'
	| 'eip155:sca:erc4337'
	| 'bip122:p2pk'
	| 'bip122:p2pkh'
	| 'bip122:p2sh'
	| 'bip122:p2wpkh'
	| 'bip122:p2tr';
```.
Notice they are all on mainnet while still differing

File: SIPS/sip-27.md (hunk 6 above)
Comment (Mrtenz) [reply SIPS/sip-27.md:55]: I feel like this just complicates things. The Snap has no information about how the address was created, right? Is it even relevant for the Snap?

File: SIPS/sip-27.md (hunk 4 above)
Comment (danroc) [reply SIPS/sip-27.md:56]: The main goal of `type` is to differentiate between accounts that may behave differently, such as EOAs and Smart Contract Accounts.

For multichain, we are still working on a solution that addresses two key issues:
- Storing past Bitcoin addresses
- Supporting multiple addresses for multiple chains (e.g., Cosmos)

We are considering approaches like `addresses: Map<Caip2ChainId, Address[]>` or `addresses: Caip10Address[]` [(PR draft](https://github.com/MetaMask/keyring-api/pull/343/files)). Which is aligned with what you said above.

Comment (danroc) [reply SIPS/sip-27.md:56]: @ritave, I need to double-check, but the reason we dropped the chains field was that CAIP-2 doesn't support wildcards. We considered using `eip155:*`, even though it wouldn’t be compliant, but ultimately decided to replace it with a `filterAccountChains` method. This method can be called with a list of CAIP-2 chain IDs, and the Snap will return the list of supported chains from the input where the account can be used.

I'm not super happy with the decision and think we can reconsided the `chains` field.

**Comment by danroc** [comment]
I'll give it a deeper read, but here are a few quick comments:

- We already have an internal account type that includes a `metadata` [field](https://github.com/MetaMask/accounts/blob/1c406e7e5052210a89eae20081f3d9aede1924b8/packages/keyring-api/src/internal/types.ts#L11-L28), but we deliberately chose not to expose that metadata to the Snaps, following the need-to-know principle.

- In the case of the send flow, I think it would make more sense to remove the burden of filtering the accounts from the Snap and have the component handle it based on a prop---something like `<AccountsList include="chainId:bip122:000000000019d6689c085ae165831e93" />`.

- The approach above also eliminates the need to expose all accounts to the Snap, which aligns with the first point.

- The Keyring API defines an API implemented by the Snap. For example, MetaMask or the companion dapp can call the `keyring_listAccounts` method to list the accounts managed by the Snap. From the usage of this new proposed method, it seems more like a method that should be exposed by the wallet to be called by the Snap, rather than being part of the Keyring API.

File: SIPS/sip-27.md (hunk 4 above)
Comment (ccharly) [reply SIPS/sip-27.md:56]: IIRC, we were also considering using just the CAIP namespace as an alternative to wildcard. This way we can regroup "things" on the same namespace, so instead of `eip155:*`, we would use `eip155`.

We started using this pattern in the extension/core IIRC. But this is still not 100% compliant with a true CAIP-2 identifier...

File: SIPS/sip-27.md [hunk 7]
```diff
@@ -0,0 +1,70 @@
...
+  address: string; // A blockchain specific public address for the account.
+  name: string; // User-given nickname for the account in the extension
+  type: string; // Blockchain specific type of the account. For example "eip155:erc4337"
+};
```
ccharly, line 59: If we wish to have a similar UI than the native one to list the accounts, we might needs some more information coming from the `metadata` field from `InternalAccount` (maybe not everything, but I was thinking about hardware-wallet accounts).

In case of HW, the `account.type` will be `eip155:eoa`, but we don't encode any other information regarding which HW if comes from. We do have this information in the `account.metadata.keyring.type` field like:

```json
{
  "id": "651e5769-86b1-4717-ae7b-63a00d7b3872",
  "address": "0x9f753769a4b45e8e2ed61108f1174b5a0939e08e",
  "options": {},
  "methods": [
    ...
  ],
  "type": "eip155:eoa",
  "metadata": {
    "name": "Ledger 1",
    "importTime": 1727081928281,
    "keyring": {
      "type": "Ledger Hardware"
    },
    "lastSelected": 1727081928285,
    "nameLastUpdatedAt": 1727081928286
  }
}
```

And I believe, if we implement a similar `AccountList` component to display the list of accounts, we will need this `keyring.type` information to be able to display the small pill `Ledger` alongside the account entry.

Screenshot 2024-09-23 at 11 16 55


But I think @danroc had a different ideas regarding rendering a `AccountList` component, so maybe we won't even need to expose any of those `metadata` :).

I just wanted to be explicit here that `type` lacks some context/information.

eriknson:
> * In the case of the send flow, I think it would make more sense to remove the burden of filtering the accounts from the Snap and have the component handle it based on a prop---something like `<AccountsList include="chainId:bip122:000000000019d6689c085ae165831e93" />`.
> * The approach above also eliminates the need to expose all accounts to the Snap, which aligns with the first point.

Created [an issue](https://github.com/MetaMask/snaps/issues/2752) to start definition of this component here, feel free to drop ideas, requirements, and thoughts etc. directly on there

SIPS/sip-27.md
ritave: This provides a stable identity for an account.

For example, you could want to map accounts to a user editable notes about those accounts.

You can't use the name since it's user editable and can change, and you can't use address because there are multiple accounts with the same address.

By providing a stable identity (similar to React's key property) you can easily manipulate those accounts.

PR #150, opened 2024-10-09
SIP-28: Background Events

ritave:
This PR proposes a SIP that allows a developer to schedule one-off events in the future.

### DIFF
--- a/SIPS/sip-28.md
+++ b/SIPS/sip-28.md
@@ -8,11 +8,11 @@ created: 2024-10-09
 
 ## Abstract
 
-This SIP introduces a way the schedule non-recurring events in the future.
+This SIP introduces a way to schedule non-recurring events in the future.
 
 ## Motivation
 
-Scheduled recurring events are already support in Snaps through Cronjobs feature. Introducing non-recurring events we allow novel use-cases for Snap developers, such as allowing a snap that sets a reminder for ENS domain expiration date.
+Scheduled recurring events are already supported in the Snaps platform via the Cronjobs feature. By introducing non-recurring events we will allow novel use-cases for Snap developers, such as allowing a snap that sets a reminder for ENS domain expiration date.
 
 ## Specification
 
@@ -26,7 +26,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
 
 ### Snap Manifest
 
-This SIP introduces a new permission `endowment:background-events`. This permissions grants a Snap the ability to schedule future events.
+This SIP introduces a new permission `endowment:background-events`. This permission grants a Snap the ability to schedule future events.
 
 This permission takes no parameters and is specified in the `snap.manifest.json` as follows:
 
@@ -44,6 +44,13 @@ This permission takes no parameters and is specified in the `snap.manifest.json`
 
 This method allows a Snap to schedule a callback to `onBackgroundEvent` handler in the future with a JSON-RPC request object as a parameter.
 
+```typescript
+type ScheduleBackgroundEventParams = {
+  date: string;
+  request: Json;
+};
+```
+
 The RPC method takes two parameters:
 
 - `date` - An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date and time and optional timezone offset.
@@ -52,9 +59,11 @@ The RPC method takes two parameters:
     > Use ISO's `Z` identifier if you want to use UTC time.
 - `request` - A JSON object that will provided as-is to `onBackgroundEvent` handler as parameter.
 
+An example of usage is given below.
+
 ```typescript
 snap.request({
-  method: "snap_backgroundEventSchedule",
+  method: "snap_scheduleBackgroundEvent",
   params: {
     date: "2024-10-09T09:59",
     request: {
@@ -73,13 +82,19 @@ The RPC method call returns a `string` that is a unique ID representing that spe
 
 This method allows to cancel an already scheduled background event using the unique ID returned from `snap_backgroundEventSchedule`
 
+```typescript
+type CancelBackgroundEventParams = { id: string };
+```
+
 This RPC method takes one argument:
 
 - `id` - The id that was returned during schedule.
 
+An example of usage is given below.
+
 ```typescript
 snap.request({
-  method: "snap_backgroundEventCancel",
+  method: "snap_cancelBackgroundEvent",
   params: {
     id: myReturnedId,
   },
@@ -88,7 +103,11 @@ snap.request({
 
 ### `onBackgroundEvent` handler
 
-This SIP introduced a new handler called `onBackgroundEvent` which is called when a scheduled background event occurs.
+This SIP introduces a new handler called `onBackgroundEvent` which is called when a scheduled background event occurs.
+
+```typescript
+type OnBackgroundEventHandler = (args: { request: Json }) => Promise<void>;
+```
 
 It has one parameter - `request`, which SHALL be provided without change from the one given as a parameter to `snap_scheduleBackgroundEvent`.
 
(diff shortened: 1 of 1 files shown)


SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+{
+  "initialPermissions": {
+    "endowment:background-event": {}
+  }
+}
+```
+
+### RPC Methods
+
+#### `snap_backgroundEventSchedule`
```
Mrtenz, line 45:
```suggestion
#### `snap_scheduleBackgroundEvent`
```

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+        foo: "bar",
+      },
+    },
+  },
+});
+```
+
+The RPC method call returns a `string` that is a unique ID representing that specific background event, allowing it to be cancelled.
+
+#### `snap_backgroundEventCancel`
```
Mrtenz, line 72:
```suggestion
#### `snap_cancelBackgroundEvent`
```

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+### Snap Manifest
+
+This SIP introduces a new permission `endowment:background-event`. This permissions grants a Snap the ability to schedule future events.
+
+This permission takes no parameters and is specified in the `snap.manifest.json` as follows:
+
+```json
+{
+  "initialPermissions": {
+    "endowment:background-event": {}
```
Mrtenz, line 38: Pluralise maybe?
```suggestion
    "endowment:background-events": {}
```

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+    id: myReturnedId,
+  },
+});
+```
+
+### `onBackgroundEvent` handler
+
+This SIP introduced a new handler called `onBackgroundEvent` which is called when a scheduled background event occurs.
+
+It has one parameter - `request`, that is provided without change.
```
Mrtenz, line 93: What do you mean by this?

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+
+### RPC Methods
+
+#### `snap_backgroundEventSchedule`
+
+This method allows a Snap to schedule a callback to `onBackgroundEvent` handler in the future with a JSON-RPC request object as a parameter.
+
+The RPC method takes two parameters:
+
+- `date` - An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date and time and optional timezone offset.
```
FrederikBolding, line 51: Do we assume UTC or local time if no timezone is provided?

FrederikBolding, line 51: Is `date` preferred to letting the caller specify a number of seconds or something?

Wondering what is the simplest for developers.

ritave: In ISO8601 if no time-zone is provided, it's local time, if there's a `Z` timezone, it means UTC.

`2024-10-09T09:59:38Z` vs `2024-10-09T09:59:38`

FrederikBolding: Ah, makes sense!

SIPS/sip-28.md
ritave: Added a comment

SIPS/sip-28.md
ritave: * Notice that we have minute rather than second level precision.
* Conversion to ISO strings is a single call to `date.toISOString()`.
* What happens to seconds when daylight time change happens?
* What happens to seconds when user travels between timezones (local-time vs UTC distinction is an edge case here, where local-time changes while UTC one doesn't)

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+sip: 28
+title: Background Events
+status: Draft
+author: Olaf Tomalka (@ritave)
+created: 2024-10-09
+---
+
+## Abstract
+
+This SIP introduces a way the schedule non-recurring events in the future.
```
FrederikBolding, line 11:
```suggestion
This SIP introduces a way to  schedule non-recurring events in the future.
```

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Motivation
+
+Scheduled recurring events are already support in Snaps through Cronjobs feature. Introducing non-recurring events we allow novel use-cases for Snap developers, such as allowing a snap that sets a reminder for ENS domain expiration date.
```
FrederikBolding, line 15:
```suggestion
Scheduled recurring events are already supported in the Snaps platform via the Cronjobs feature. By introducing non-recurring events we will allow novel use-cases for Snap developers, such as allowing a snap that sets a reminder for ENS domain expiration date.
```

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Snap Manifest
+
+This SIP introduces a new permission `endowment:background-events`. This permissions grants a Snap the ability to schedule future events.
```
FrederikBolding, line 29:
```suggestion
This SIP introduces a new permission `endowment:background-events`. This permission grants a Snap the ability to schedule future events.
```

FrederikBolding, line 29: Wondering if we should somehow merge this with cronjobs 🤔 Otherwise, we'll have two permissions that do extremely similar things 😄

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+  method: "snap_backgroundEventCancel",
+  params: {
+    id: myReturnedId,
+  },
+});
+```
+
+### `onBackgroundEvent` handler
+
+This SIP introduced a new handler called `onBackgroundEvent` which is called when a scheduled background event occurs.
```
Mrtenz, line 91:
```suggestion
This SIP introduces a new handler called `onBackgroundEvent` which is called when a scheduled background event occurs.
```

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+```typescript
+snap.request({
+  method: "snap_backgroundEventCancel",
+  params: {
+    id: myReturnedId,
+  },
+});
+```
+
+### `onBackgroundEvent` handler
```
Mrtenz, line 89: Please add a TypeScript definition for this.

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+        foo: "bar",
+      },
+    },
+  },
+});
+```
+
+The RPC method call returns a `string` that is a unique ID representing that specific background event, allowing it to be cancelled.
+
+#### `snap_cancelBackgroundEvent`
```
Mrtenz, line 72: Please add a TypeScript definition for this.

SIPS/sip-28.md
```diff
@@ -0,0 +1,97 @@
...
+{
+  "initialPermissions": {
+    "endowment:background-events": {}
+  }
+}
+```
+
+### RPC Methods
+
+#### `snap_scheduleBackgroundEvent`
```
Mrtenz, line 43: Please add a TypeScript definition for this.

SIPS/sip-28.md
FrederikBolding: WDYT @Mrtenz?

FrederikBolding: I think we eventually want to rename cronjobs to background events, but doing so will be a breaking change.

FrederikBolding (approved):

Mrtenz (approved):

SIPS/sip-28.md
legobeat: I think that defaulting to localtime (which may not always be canonically known) can become problematic and would like to propose that there should be no ambiguity in interpretation:
- #152

PR #152, opened 2024-10-15
SIP-28: Omitted timezone default to non-ambiguous UTC

legobeat:


hmalik88 (approved):
Yeah, I think I agree here. We should treat omitted timezone as UTC and all together not require the `Z` identifier.

legobeat:
Something to consider would be to make the timezone required - this would reduce the likelihood of bugs arising due to incorrect implementations naively using (non-UTC) defaults in some library other than UTC, at the cost of making the `Z` required for UTC.

PR #154, opened 2024-12-05
SIP-29: Snap Assets API

danroc:
This SIP aims to define a new API that can be exposed by Snaps to allow clients to retrieve asset information in a chain-agnostic way.

### DIFF
--- a/SIPS/sip-29.md
+++ b/SIPS/sip-29.md
@@ -1,0 +1,196 @@
+---
+sip: (To be assigned)
+title: Snap Assets API
+status: Draft
+discussions-to (*optional): (Http/Https URL)
+author: Daniel Rocha (@danroc), Guillaume Roux (@GuillaumeRx)
+created: (Date created on in ISO 8601 format. `yyyy-mm-dd`)
+updated (*optional): (Date last updated on in https://en.wikipedia.org/wiki/ISO_8601 format. `yyyy-mm-dd`. This should be only used on SIPs with `Living` status)
+---
+
+## Abstract
+
+This SIP aims to define a new API that can be exposed by Snaps to allow clients
+get asset information in a chain-agnostic way.
+
+## Motivation
+
+To allow clients to be chain-agnostic, the logic of how to get asset
+information should be abstracted away from the client. We also need to define
+the types that represent the asset information required by the clients.
+
+## Specification
+
+> Indented sections like this are considered non-normative.
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" written
+in uppercase in this document are to be interpreted as described in [RFC
+2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Definitions
+
+In this document, all definitions are written in TypeScript. Also, any time an
+asset need to be identified, it MUST use the [CAIP-19][caip-19] representation.
+
+### Snap Assets API
+
+Two methods are defined in the Snap Assets API:
+
+#### Get Token Description
+
+```typescript
+// Represents a token unit.
+type TokenUnit {
+    // Human-friendly name of the token unit.
+    name: string;
+
+    // Ticker of the token unit.
+    ticker: string;
+
+    // Number of decimals of the token unit.
+    decimals: number;
+}
+
+// Token description.
+type TokenDescription {
+    // Human-friendly name of the token.
+    name: string;
+
+    // Ticker of the token.
+    ticker: string;
+
+    // Whether the token is native to the chain.
+    isNative: boolean;
+
+    // Base64 representation of the token icon.
+    iconBase64: string;
+
+    // List of token units.
+    units: TokenUnit[];
+}
+
+// Returns the description of a non-fungible token. This description can then
+// be used by the client to display relevant information about the token.
+//
+// @example
+// ```typescript
+// const tokenDescription = await getTokenDescription('eip155:1/slip44:60');
+//
+// // Returns:
+// // {
+// //     name: 'Ether',
+// //     ticker: 'ETH',
+// //     isNative: true,
+// //     iconBase64: 'data:image/png;base64,...',
+// //     units: [
+// //         {
+// //             name: 'Ether',
+// //             ticker: 'ETH',
+// //             decimals: 18
+// //         },
+// //         {
+// //             name: 'Gwei',
+// //             ticker: 'Gwei',
+// //             decimals: 9
+// //         },
+// //         {
+// //             name: 'wei',
+// //             ticker: 'wei',
+// //             decimals: 0
+// //         }
+// //     ]
+// ```
+function getTokenDescription(token: Caip19AssetType): TokenDescription;
+```
+
+#### Get Token Conversion Rate
+
+```typescript
+type TokenConversionRate {
+    // The rate of conversion from the source token to the target token. It
+    // means that 1 unit of the `from` token should be converted to this amount
+    // of the `to` token.
+    rate: string;
+
+    // The UNIX timestamp of when the conversion rate was last updated.
+    conversionTime: number;
+
+    // The UNIX timestamp of when the conversion rate will expire.
+    expirationTime: number;
+}
+
+// Returns the conversion rate between two assets (tokens or fiat).
+//
+// @example
+// ```typescript
+// const conversionRate = await getTokenConversionRate(
+//   'eip155:1/slip44:60',
+//   'eip155:1/erc20:0x6b175474e89094c44da98b954eedeac495271d0f'
+// );
+//
+// // Returns:
+// // {
+// //     rate: '3906.38',
+// //     conversionTime: 1733389786,
+// //     expirationTime: 1733389816,
+// // }
+// ```
+function getTokenConversionRate(
+    from: Caip19AssetType,
+    to: Caip19AssetType
+): TokenConversionRate;
+```
+
+### Fiat currency representation
+
+We SHOULD use CAIP-19 to represent fiat currencies as well, it will allow us to
+have a consistent way to represent all assets and make the API more
+predictable.
+
+The propose format is:
+
+```
+asset_type:        chain_id + "/" + asset_namespace + ":" + asset_reference
+chain_id:          namespace + ":" + reference
+namespace:         "fiat"
+reference:         country_code
+asset_namespace:   "currency"
+asset_reference:   currency_code
+```
+
+The country code is a two-letter lowercase code as defined by the ISO 3166-1
+alpha-2, with the exception for the European Union, which is represented by
+"eu".
+
+The currency code is a three-letter uppercase code as defined by the ISO 4217.
+
+Examples:
+
+```
+# Euro
+fiat:eu/currency:eur
+
+# United States Dollar
+fiat:us/currency:usd
+
+# Brazilian Real
+fiat:br/currency:brl
+
+# Japanese Yen
+fiat:jp/currency:jpy
+```
+
+## Backwards compatibility
+
+Any SIPs that break backwards compatibility MUST include a section describing
+those incompatibilities and their severity. The SIP SHOULD describe how the
+author plans on proposes to deal with such these incompatibilities.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE).
+
+[caip-19]: https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-19.md
--- a/SIPS/sip-29.md
+++ b/SIPS/sip-29.md
@@ -144,7 +144,7 @@ type OnAssetConversionReturn = {
 
 ### Fiat currency representation
 
-We SHOULD use CAIP-19 to represent fiat currencies as well. This approach
+We SHOULD use [CAIP-19][caip-19] to represent fiat currencies as well. This approach
 provides a consistent way to represent all assets, making the API more
 predictable. The proposed format is:
 
--- a/SIPS/sip-29.md
+++ b/SIPS/sip-29.md
@@ -33,7 +33,7 @@ in uppercase in this document are to be interpreted as described in [RFC
 1. In this document, all definitions are written in TypeScript.
 
 2. Any time an asset needs to be identified, it MUST use the [CAIP-19][caip-19]
-representation.
+   representation.
 
 ### Snap Manifest
 
@@ -46,9 +46,9 @@ This permission is specified as follows in `snap.manifest.json` files:
 {
   "initialPermissions": {
     "endowment:assets": {
-        "scopes": [
-            "bip122:000000000019d6689c085ae165831e93"
-        ]
+      "scopes": [
+        "bip122:000000000019d6689c085ae165831e93"
+      ]
     }
   }
 }
@@ -77,17 +77,17 @@ export const onAssetLookup: OnAssetLookupHandler = async ({
 
 The type for an `onAssetLookup` handler function’s arguments is:
 
-
 ```typescript
 interface OnAssetLookupArgs {
-    assets: Caip19AssetType[];
+  assets: Caip19AssetType[];
 }
 ```
+
 The type for an `onAssetLookup` handler function’s return value is:
 
 ```typescript
 type OnAssetLookupReturn = {
-    assets: Record<Caip19AssetType, AssetMetadata>;
+  assets: Record<Caip19AssetType, AssetMetadata>;
 };
 ```
 
@@ -103,32 +103,34 @@ export const onAssetConversion: OnAssetConversionHandler = async ({
   return { conversionRates };
 };
 ```
+
 The type for an `onAssetConversion` handler function’s arguments is:
 
 ```typescript
 type Conversion = {
-    from: Caip19AssetType;
-    to: Caip19AssetType;
+  from: Caip19AssetType;
+  to: Caip19AssetType;
 };
 
 type OnAssetConversionArgs = {
-    conversions: Conversion[];
-}
+  conversions: Conversion[];
+};
 ```
+
 The type for an `onAssetConversion` handler function’s return value is:
 
 ```typescript
 type AssetConversionRate = {
-    // The rate of conversion from the source asset to the target asset. It
-    // means that 1 unit of the `from` asset should be converted to this amount
-    // of the `to` asset.
-    rate: string;
+  // The rate of conversion from the source asset to the target asset. It
+  // means that 1 unit of the `from` asset should be converted to this amount
+  // of the `to` asset.
+  rate: string;
 
-    // The UNIX timestamp of when the conversion rate was last updated.
-    conversionTime: number;
+  // The UNIX timestamp of when the conversion rate was last updated.
+  conversionTime: number;
 
-    // The UNIX timestamp of when the conversion rate will expire.
-    expirationTime: number;
+  // The UNIX timestamp of when the conversion rate will expire.
+  expirationTime: number;
 };
 
 type FromAsset = Conversion["from"];
@@ -136,7 +138,7 @@ type FromAsset = Conversion["from"];
 type ToAsset = Conversion["to"];
 
 type OnAssetConversionReturn = {
-    conversionRates: Record<From, Record<To, AssetConversionRate>>;
+  conversionRates: Record<From, Record<To, AssetConversionRate>>;
 };
 ```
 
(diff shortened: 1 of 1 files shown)


SIPS/sip-29.md
```diff
@@ -0,0 +1,195 @@
...
+// //     expirationTime: 1733389816,
+// // }
+// ```
+function getTokenConversionRate(
+    from: Caip19AssetType,
+    to: Caip19AssetType
+): TokenConversionRate;
+```
+
+### Fiat currency representation
```
GuillaumeRx, line 148: We should create a CAIP for this

SIPS/sip-29.md
```diff
@@ -0,0 +1,195 @@
...
+
+    // Ticker of the token unit.
+    ticker: string;
+
+    // Number of decimals of the token unit.
+    decimals: number;
+};
+
+// Token description.
+type TokenDescription = {
```
GuillaumeRx, line 58: We should probably keep a reference to the `CAIP-19` id here

SIPS/sip-29.md
danroc: This is a good point. This is technically CAIP-19, but with unregistered namespaces.

I started a [discussion](https://github.com/ChainAgnostic/CAIPs/discussions/338) on the CAIPs repo, let's see what other think of it.

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+  ticker: string;
+
+  // Whether the asset is native to the chain.
+  native: boolean;
+
+  // Represents a fungible asset
+  fungible: true;
+
+  // Base64 representation of the asset icon.
+  iconBase64: string;
```
danroc, line 215: > Can this be an URL link to the icon?

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+// Fungible asset metadata.
+type FungibleAssetMetadata = {
+  // Human-friendly name of the asset.
+  name: string;
+
+  // Ticker of the asset.
+  ticker: string;
+
+  // Whether the asset is native to the chain.
+  native: boolean;
```
danroc, line 209: > How can we identify other types of tokens that need to be handled differently?

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+
+type OnAssetConversionArgs = {
+  conversions: Conversion[];
+};
+```
+
+The type for an `onAssetConversion` handler function’s return value is:
+
+```typescript
+type AssetConversionRate = {
```
danroc, line 123: > How can we get historical market data?

SIPS/sip-29.md
bergeron: Some examples might be differentiating between brc-20 and rune tokens on bitcoin. Or eth, erc20s, and staking contracts on ethereum.  I anticipate more types of tokens than just native vs not. Not sure if anything built into caip19 already helps distinguish this.

SIPS/sip-29.md
aganglada: I think it has to be an icon, the static token icon server is going to return pngs

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+```typescript
+interface OnAssetLookupArgs {
+  assets: Caip19AssetType[];
+}
+```
+
+The type for an `onAssetLookup` handler function’s return value is:
+
+```typescript
+type OnAssetLookupReturn = {
```
FrederikBolding, line 89:
```suggestion
type OnAssetLookupResponse = {
```

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+
+  // The UNIX timestamp of when the conversion rate will expire.
+  expirationTime: number;
+};
+
+type FromAsset = Conversion["from"];
+
+type ToAsset = Conversion["to"];
+
+type OnAssetConversionReturn = {
```
FrederikBolding, line 140:
```suggestion
type OnAssetConversionResponse = {
```

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+  // The rate of conversion from the source asset to the target asset. It
+  // means that 1 unit of the `from` asset should be converted to this amount
+  // of the `to` asset.
+  rate: string;
+
+  // The UNIX timestamp of when the conversion rate was last updated.
+  conversionTime: number;
+
+  // The UNIX timestamp of when the conversion rate will expire.
+  expirationTime: number;
```
salimtb, line 133: hey @danroc 
The data we currently have does not include MarketData, which means we cannot display the 1-day price changes or chart-based price changes. I believe this should be updated to include an interface like the following:
```
interface MarketData {
	id: string;
	price: number;
	marketCap: number;
	allTimeHigh: number;
	allTimeLow: number;
	totalVolume: number;
	high1d: number;
	circulatingSupply: number;
	dilutedmarketCap: number;
	marketCapPercentChange1d: number;
	priceChange1d: number;
	pricePercentChange1h: number;
	pricePercentChange1d: number;
	pricePercentChange7d: number;
	pricePercentChange14d: number;
	pricePercentChange30d: number;
	pricePercentChange200d: number;
	pricePercentChange1y: number;
}
```

danroc: Hey @salimtb, thanks! We didn’t plan to include market data in the first release, so we left it out of this API to give ourselves more time to figure out what’s really needed.

Are there any fields you’d like to add or remove from the current EVM implementation? Now might be a good time to make any changes you’ve been thinking about.

Also, it’s not a big deal if it gets added later to this SIP or even a new one.

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+```typescript
+type OnAssetLookupReturn = {
+  assets: Record<Caip19AssetType, AssetMetadata>;
+};
+```
+
+#### Get Asset Conversion Rate
+
+```typescript
+import { OnAssetConversionHandler } from "@metamask/snaps-sdk";
```
salimtb, line 97: is this function implemented within the @metamask/snap-sdk ? is it part of the assets team scope ?

SIPS/sip-29.md
salimtb: thanks for your response, @danroc. Not all the fields are currently in use, but we decided to keep them for potential future features. If you'd like, I can provide you with a list of the properties that are currently being used.
all of the properties are returned by the endpoint `/spot-price` of the price API

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+    "endowment:assets": {
+      "scopes": [
+        "bip122:000000000019d6689c085ae165831e93"
+      ]
+    }
+  }
+}
+```
+
+`scopes` - A non-empty array of CAIP-2 chain IDs that the snap supports. This field is useful for a client in order to avoid unnecessary overhead.
```
ccharly, line 57:
```suggestion
`scopes` - A non-empty array of CAIP-2 chain IDs that the Snap supports. This field is useful for a client in order to avoid unnecessary overhead.
```

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+interface OnAssetLookupArgs {
+  assets: Caip19AssetType[];
+}
+```
+
+The type for an `onAssetLookup` handler function’s return value is:
+
+```typescript
+type OnAssetLookupReturn = {
+  assets: Record<Caip19AssetType, AssetMetadata>;
```
ccharly, line 90: Is the `AssetMetadata` type missing? Or should we make a reference to the `FungibleAssetMetadata`?

Could also be `type AssetMetadata = FungibleAssetMetadata` for now (if we omit the NFT support).

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+
+#### Get Asset Metadata
+
+```typescript
+import { OnAssetLookupHandler } from "@metamask/snaps-sdk";
+
+export const onAssetLookup: OnAssetLookupHandler = async ({
+  assets
+}) => {
+  const assetsMetadata = /* Get metadata */;
```
ccharly, line 73: Nit: Just being explicit here:

```suggestion
  const assetsMetadata = /* Get metadata for given `assets` */;
```

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+
+#### Get Asset Conversion Rate
+
+```typescript
+import { OnAssetConversionHandler } from "@metamask/snaps-sdk";
+
+export const onAssetConversion: OnAssetConversionHandler = async ({
+  conversions
+}) => {
+  const conversionRates = /* Get conversion rate */;
```
ccharly, line 102: Nit:

```suggestion
  const conversionRates = /* Get conversion rate for given `conversions` */;
```

SIPS/sip-29.md
ccharly: Nit: Do we want to limit the size of this icon somehow?

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+type ToAsset = Conversion["to"];
+
+type OnAssetConversionReturn = {
+  conversionRates: Record<From, Record<To, AssetConversionRate>>;
+};
+```
+
+### Fiat currency representation
+
+We SHOULD use CAIP-19 to represent fiat currencies as well. This approach
```
ccharly, line 147:
```suggestion
We SHOULD use [CAIP-19][caip-19] to represent fiat currencies as well. This approach
```

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+}
+```
+
+`scopes` - A non-empty array of CAIP-2 chain IDs that the snap supports. This field is useful for a client in order to avoid unnecessary overhead.
+
+### Snap Implementation
+
+Two methods are defined in the Snap Assets API:
+
+Any Snap that wishes to provide asset information **MUST** implement the following API:
```
ccharly, line 63: Very nit: Other SIPs do not seem to use formatting for those keywords, so maybe we should remove it here too?

```suggestion
Any Snap that wishes to provide asset information MUST implement the following API:
```

SIPS/sip-29.md
danroc: No, the Assets team will be a caller of this method, but its implementation will be handled by other teams (Solana and Snaps).

SIPS/sip-29.md
danroc: Indeed, we forgot to add the type alias when we renamed the type from `AssetMetadata` to `FungibleAssetMetadata`.

SIPS/sip-29.md
```diff
@@ -0,0 +1,232 @@
...
+    "endowment:assets": {
+      "scopes": [
+        "bip122:000000000019d6689c085ae165831e93"
+      ]
+    }
+  }
+}
+```
+
+`scopes` - A non-empty array of CAIP-2 chain IDs that the Snap supports. This field is useful for a client in order to avoid unnecessary overhead.
```
ccharly, line 57: > to avoid unnecessary overhead.

Since this information would be static, will the Snap only be "restricted" to those scopes? Or can Snap also returns assets informations for other non-declared scopes?

FrederikBolding: My expectation for this would be that on the controller end we would only query this Snap for assets if we are interested in the scopes that it has.

SIPS/sip-29.md
FrederikBolding: Is this in scope?

SIPS/sip-29.md
FrederikBolding: What's the status of this discussion?

SIPS/sip-29.md
```diff
@@ -0,0 +1,235 @@
...
+
+  // Base64 representation of the asset icon.
+  iconBase64: string;
+
+  // List of asset units.
+  units: FungibleAssetUnit[];
+};
+
+// Represents the metadata of an asset.
+type AssetMetadata = FungibleAssetMetadata
```
FrederikBolding, line 222: Are these types already defined somewhere for the keyring API part of this?

SIPS/sip-29.md
```diff
@@ -0,0 +1,235 @@
...
+}) => {
+  const assetsMetadata = /* Get metadata for given `assets` */;
+  return { assets: assetsMetadata };
+};
+```
+
+The type for an `onAssetLookup` handler function’s arguments is:
+
+```typescript
+interface OnAssetLookupArgs {
```
FrederikBolding, line 81:
```suggestion
interface OnAssetLookupArguments {
```

SIPS/sip-29.md
```diff
@@ -0,0 +1,235 @@
...
+
+The type for an `onAssetConversion` handler function’s arguments is:
+
+```typescript
+type Conversion = {
+  from: Caip19AssetType;
+  to: Caip19AssetType;
+};
+
+type OnAssetConversionArgs = {
```
FrederikBolding, line 115:
```suggestion
type OnAssetConversionArguments = {
```

SIPS/sip-29.md
```diff
@@ -0,0 +1,235 @@
...
+Two methods are defined in the Snap Assets API:
+
+Any Snap that wishes to provide asset information MUST implement the following API:
+
+#### Get Asset Metadata
+
+```typescript
+import { OnAssetLookupHandler } from "@metamask/snaps-sdk";
+
+export const onAssetLookup: OnAssetLookupHandler = async ({
```
Mrtenz, line 70: Maybe pluralise to match the permission name.

```suggestion
export const onAssetsLookup: OnAssetsLookupHandler = async ({
```

Mrtenz (changes requested):

SIPS/sip-29.md
FrederikBolding: In that case, should we update the other handler too? 🤔

SIPS/sip-29.md
```diff
@@ -0,0 +1,235 @@
...
+  const assetsMetadata = /* Get metadata for given `assets` */;
+  return { assets: assetsMetadata };
+};
+```
+
+The type for an `onAssetLookup` handler function’s arguments is:
+
+```typescript
+interface OnAssetLookupArgs {
+  assets: Caip19AssetType[];
```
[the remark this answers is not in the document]
FrederikBolding: That's what the CAIP-19 identifier is also called: https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-19.md#specification-of-asset-type

SIPS/sip-29.md
Mrtenz: Yes, it should be consistent.

SIPS/sip-29.md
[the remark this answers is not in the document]
Mrtenz: That should be defined/clarified somewhere.

SIPS/sip-29.md
danroc: Not for v1. We might want to add it in the near future tho.

SIPS/sip-29.md
```diff
@@ -0,0 +1,235 @@
...
+  symbol: string;
+
+  // Whether the asset is native to the chain.
+  native: boolean;
+
+  // Represents a fungible asset
+  fungible: true;
+
+  // Base64 representation of the asset icon.
+  iconBase64: string;
```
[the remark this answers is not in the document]
danroc: @aganglada do you know which format is used by the "Icons API"?

SIPS/sip-29.md
danroc: @FrederikBolding, do you know if we have this use case in other APIs? If yes, what size limit did we use?

SIPS/sip-29.md
danroc: It didn't move further. @GuillaumeRx, do you think you can check with CASA again?

SIPS/sip-29.md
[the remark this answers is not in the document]
aganglada: Image URLs

https://www.notion.so/metamask-consensys/Multichain-Initiative-wip-New-Chain-Agnostic-Endpoints-151f86d67d68809b8519e3841ea54061

SIPS/sip-29.md
GuillaumeRx: Maybe it's good to bring @hmalik88 here.

SIPS/sip-29.md
GuillaumeRx: I agree

SIPS/sip-29.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+The type for an `onAssetConversion` handler function’s return value is:
+
+```typescript
+type AssetConversionRate = {
+  // The rate of conversion from the source asset to the target asset. It
+  // means that 1 unit of the `from` asset should be converted to this amount
+  // of the `to` asset.
+  rate: string;
```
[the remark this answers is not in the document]
FrederikBolding: I'm assuming a regular decimal number, but we should clarify in the spec

SIPS/sip-29.md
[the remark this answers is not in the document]
FrederikBolding: Seems like we should change this to be `iconUrl` then. We'll need to validate these beyond just checking that they are a string for safety.

SIPS/sip-29.md
FrederikBolding: Maybe worth removing this section from the spec for now and pursue it in parallel?

SIPS/sip-29.md
FrederikBolding: But if these are going to be URLs then we likely want to validate the URLs extensively instead of the content

SIPS/sip-29.md
```diff
@@ -0,0 +1,229 @@
...
+  symbol: string;
+
+  // Whether the asset is native to the chain.
+  native: boolean;
+
+  // Represents a fungible asset
+  fungible: true;
+
+  // Base64 representation of the asset icon.
+  iconBase64: string;
```
FrederikBolding, line 215:
```suggestion
  // URL of the asset icon.
  iconUrl: string;
```

Based on conversation in other threads on this PR

SIPS/sip-29.md
```diff
@@ -0,0 +1,235 @@
...
+  // The rate of conversion from the source asset to the target asset. It
+  // means that 1 unit of the `from` asset should be converted to this amount
+  // of the `to` asset.
+  rate: string;
+
+  // The UNIX timestamp of when the conversion rate was last updated.
+  conversionTime: number;
+
+  // The UNIX timestamp of when the conversion rate will expire.
+  expirationTime: number;
```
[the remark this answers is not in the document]
FrederikBolding: If not, we could make it optional and have a reasonable fallback on our end

SIPS/sip-29.md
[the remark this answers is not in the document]
FrederikBolding: At the very least we should preserve some way to pass local images so that Snaps can display assets without requiring a CDN

SIPS/sip-29.md
GuillaumeRx: Are we doing URLs or base64 strings ? We need to choose. IIRC the API is going to return a link, do we want to download the icon in the snap ?

FrederikBolding: I think we'll need to support both: https://github.com/MetaMask/SIPs/pull/154#discussion_r1923708203

danroc: Agreed, we have the use case where Snaps will support only one asset (like Bitcoin), so it makes sense to allow the Snap to embed the image. For other chains, it makes sense to use URLs.

One proposed approach is to rename this property to `iconUrl` and allow [data URLs](https://developer.mozilla.org/en-US/docs/Web/URI/Schemes/data).

SIPS/sip-29.md
danroc: Indeed we should enforce that the Snap can return info only for assets under its scope. (This is an implementation detail of the controller.)

SIPS/sip-29.md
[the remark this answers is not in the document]
danroc: We can make it optional and assume that the `MultichainRatesController` will define default fallbacks.

SIPS/sip-29.md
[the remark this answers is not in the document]
danroc: [Related](https://github.com/MetaMask/SIPs/pull/154/files#r1923953825)

- For external images (`https:` URLs): any image
- Embedded images (`data:` URLs): SVG only (but this can be changed if needed in the future)

SIPS/sip-29.md
```diff
@@ -0,0 +1,194 @@
...
+// Fungible asset metadata.
+type FungibleAssetMetadata = {
+  // Human-friendly name of the asset.
+  name: string;
+
+  // Ticker symbol of the asset's main unit.
+  symbol: string;
+
+  // Whether the asset is native to the chain.
+  native: boolean;
```
FrederikBolding, line 174:
```suggestion
```

SIPS/sip-29.md
danroc: We can remove this field and use the asset namespace to determine its type.

Note that the [`slip44`](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-20.md) namespace is used for native tokens, which means that this field can be removed.

SIPS/sip-29.md
```diff
@@ -0,0 +1,191 @@
...
+    "endowment:assets": {
+      "scopes": [
+        "bip122:000000000019d6689c085ae165831e93"
+      ]
+    }
+  }
+}
+```
+
+`scopes` - A non-empty array of CAIP-2 chain IDs that the Snap supports. This field is useful for a client in order to avoid unnecessary overhead. Any asset returned by the Snap must be from one of the supported chains otherwise it will get filtered out.
```
FrederikBolding, line 57:
```suggestion
`scopes` - A non-empty array of CAIP-2 chain IDs that the Snap supports. This field is useful for a client in order to avoid unnecessary overhead. Any asset returned by the Snap MUST filtered out if it isn't part of the supported scopes.
```

SIPS/sip-29.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Snap Implementation
+
+Two methods are defined in the Snap Assets API:
+
+Any Snap that wishes to provide asset information MUST implement the following API:
+
+#### Get Assets Metadata
+
+`Caip19AssetType` -  A string that represents an asset using the [CAIP-19][caip-19] standard.
```
FrederikBolding, line 67:
```suggestion
`Caip19AssetType` - A string that represents an asset using the [CAIP-19][caip-19] standard.
```

SIPS/sip-29.md
```diff
@@ -0,0 +1,191 @@
...
+  // Human-friendly name of the asset.
+  name: string;
+
+  // Ticker symbol of the asset's main unit.
+  symbol: string;
+
+  // Represents a fungible asset
+  fungible: true;
+
+  // data URI or URL representation of the asset icon.
```
FrederikBolding, line 176:
```suggestion
  // Data URI or URL representation of the asset icon.
```

FrederikBolding (approved):

Mrtenz (approved):

PR #155, opened 2024-12-05
SIP-29 Improvements

GuillaumeRx:
This PR tries to add a permission and handlers definition to the `SIP-29` (https://github.com/MetaMask/SIPs/pull/154). It also propose to use `asset` rather than the `token` denomination.

### DIFF
--- a/SIPS/sip-29.md
+++ b/SIPS/sip-29.md
@@ -68,7 +68,7 @@ Any Snap that wishes to provide asset information **MUST** implement the followi
 import { OnAssetDescriptionHandler } from "@metamask/snaps-sdk";
 
 export const onAssetDescription: OnAssetDescriptionHandler = async ({
-  asset
+  assets
 }) => {
   const description = /* Get description */;
   return { description };
@@ -79,12 +79,36 @@ The type for an `onAssetDescription` handler function’s arguments is:
 
 ```typescript
 interface OnAssetDescriptionArgs {
-    asset: Caip19AssetType;
+    assets: Caip19AssetType[];
 }
 ```
 The type for an `onAssetDescription` handler function’s return value is:
 
 ```typescript
+// {
+//     name: 'Ether',
+//     ticker: 'ETH',
+//     isNative: true,
+//     iconBase64: 'data:image/png;base64,...',
+//     units: [
+//         {
+//             name: 'Ether',
+//             ticker: 'ETH',
+//             decimals: 18
+//         },
+//         {
+//             name: 'Gwei',
+//             ticker: 'Gwei',
+//             decimals: 9
+//         },
+//         {
+//             name: 'wei',
+//             ticker: 'wei',
+//             decimals: 0
+//         }
+//    ]
+// }
+
 // Represents a asset unit.
 type AssetUnit = {
     // Human-friendly name of the asset unit.
@@ -118,32 +142,9 @@ type AssetDescription = {
     units: AssetUnit[];
 };
 
-// {
-//     name: 'Ether',
-//     ticker: 'ETH',
-//     isNative: true,
-//     iconBase64: 'data:image/png;base64,...',
-//     units: [
-//         {
-//             name: 'Ether',
-//             ticker: 'ETH',
-//             decimals: 18
-//         },
-//         {
-//             name: 'Gwei',
-//             ticker: 'Gwei',
-//             decimals: 9
-//         },
-//         {
-//             name: 'wei',
-//             ticker: 'wei',
-//             decimals: 0
-//         }
-//    ]
-// }
 
 type OnAssetDescriptionReturn = {
-    description: AssetDescription;
+    description: Record<Caip19AssetType, AssetDescription>;
 };
 ```
 
@@ -164,7 +165,7 @@ The type for an `onAssetDescription` handler function’s arguments is:
 
 ```typescript
 interface OnAssetConversionArgs {
-    from: Caip19AssetType;
+    from: Caip19AssetType[];
     to: Caip19AssetType;
 }
 ```
@@ -192,7 +193,7 @@ type AssetConversionRate = {
 // }
 
 type OnAssetConversionReturn = {
-    conversionRate: AssetConversionRate;
+    conversionRate: Record<Caip19AssetType, AssetConversionRate>;
 };
 ```
 
--- a/SIPS/sip-29.md
+++ b/SIPS/sip-29.md
@@ -38,7 +38,7 @@ representation.
 ### Snap Manifest
 
 This SIP introduces a new permission named `endowment:assets`.
-This permission grants a snap the ability to provide asset information to the client.
+This permission grants a Snap the ability to provide asset information to the client.
 
 This permission is specified as follows in `snap.manifest.json` files:
 
@@ -54,7 +54,7 @@ This permission is specified as follows in `snap.manifest.json` files:
 
 Two methods are defined in the Snap Assets API:
 
-Any snap that wishes to provide asset information **MUST** implement the following API:
+Any Snap that wishes to provide asset information **MUST** implement the following API:
 
 #### Get Token Description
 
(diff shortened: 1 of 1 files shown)


SIPS/sip-29.md
```diff
@@ -35,15 +35,52 @@ in uppercase in this document are to be interpreted as described in [RFC
...
+### Snap Implementation
 
 Two methods are defined in the Snap Assets API:
 
+Any snap that wishes to provide asset information **MUST** implement the following API:
+
 #### Get Token Description
 
+```typescript
+import { OnAssetDescriptionHandler } from "@metamask/snap-types";
```
FrederikBolding, line 62:
```suggestion
import { OnAssetDescriptionHandler } from "@metamask/snaps-sdk";
```

SIPS/sip-29.md
```diff
@@ -35,15 +35,52 @@ in uppercase in this document are to be interpreted as described in [RFC
 2. Any time an asset needs to be identified, it MUST use the [CAIP-19][caip-19]
 representation.
 
-### Snap Assets API
+### Snap Manifest
+
+This SIP introduces a new permission named `endowment:assets`.
+This permission grants a snap the ability to provide asset information to the client.
```
FrederikBolding, line 41:
```suggestion
This permission grants a Snap the ability to provide asset information to the client.
```

SIPS/sip-29.md
```diff
@@ -35,15 +35,52 @@ in uppercase in this document are to be interpreted as described in [RFC
...
 Two methods are defined in the Snap Assets API:
 
+Any snap that wishes to provide asset information **MUST** implement the following API:
+
 #### Get Token Description
 
+```typescript
+import { OnAssetDescriptionHandler } from "@metamask/snap-types";
+
+export const onAssetDescription: OnAssetDescriptionHandler = async ({
```
FrederikBolding, line 64: Wondering if we need the split to have two different handlers 🤔 

Perhaps a question for the original SIP, but wondering if we could simply have a lookup function that would return the known pricing and metadata combined reducing the amount of potential calls

FrederikBolding: E.g. `onAssetLookup` or something

GuillaumeRx: @danroc what's your take on this ?

GuillaumeRx: We'll need to poll frequently for the asset rate on the client side and the asset description might be unused data in this case. If the snap doesn't cache the asset metadata it will mean that a call to an API is done everytime for nothing.

SIPS/sip-29.md
```diff
@@ -35,85 +35,140 @@ in uppercase in this document are to be interpreted as described in [RFC
...
+
+This SIP introduces a new permission named `endowment:assets`.
+This permission grants a Snap the ability to provide asset information to the client.
+
+This permission is specified as follows in `snap.manifest.json` files:
+
+```json
+{
+  "initialPermissions": {
+    "endowment:assets": {}
```
GuillaumeRx, line 48: I'm wondering if a caveat for the supported chains here would be useful.

GuillaumeRx:
```json
{
  "initialPermissions": {
    "endowment:assets": {
     "chains": [
         "foo:bar"
       ]
    }
}
```

FrederikBolding: Probably yeah - could work as a way to reduce calls to Snaps that know about assets on unwanted chains

SIPS/sip-29.md
FrederikBolding: That's a good point 🤔

danroc: Indeed, I agree with @GuillaumeRx, those are two distinct concerns and the methods are not called with same frequency.

SIPS/sip-29.md
```diff
@@ -35,85 +35,140 @@ in uppercase in this document are to be interpreted as described in [RFC
...
+};
 ```
 
-#### Get Token Conversion Rate
+#### Get Asset Conversion Rate
+
+```typescript
+import { OnAssetConversionHandler } from "@metamask/snaps-sdk";
+
+export const onAssetConversion: OnAssetConversionHandler = async ({
```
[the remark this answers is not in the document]
danroc: No objections on supporting batching 👍

danroc:
I wasn't sure whether to use asset or token in the method names, as asset is more generic and could include NFTs, for instance, where the details payload would differ significantly from that of tokens.

SIPS/sip-29.md
```diff
@@ -35,85 +35,147 @@ in uppercase in this document are to be interpreted as described in [RFC
...
+The type for an `onAssetDescription` handler function’s arguments is:
 
 ```typescript
-type TokenConversionRate = {
-    // The rate of conversion from the source token to the target token. It
-    // means that 1 unit of the `from` token should be converted to this amount
-    // of the `to` token.
+interface OnAssetConversionArgs {
+    from: Caip19AssetType[];
+    to: Caip19AssetType;
```
GuillaumeRx, line 169: I'm wondering if we should also support multiple `to`

GuillaumeRx:
> I wasn't sure whether to use `asset` or `token` in the method names, as asset is more generic and could include NFTs, for instance, where the details payload would differ significantly from that of tokens.

I wonder if this API couldn't be versatile enough that we could add the support for NFT down the road rather than having multiple endpoints. It should just be a type change

FrederikBolding:
> > I wasn't sure whether to use `asset` or `token` in the method names, as asset is more generic and could include NFTs, for instance, where the details payload would differ significantly from that of tokens.
> 
> I wonder if this API couldn't be versatile enough that we could add the support for NFT down the road rather than having multiple endpoints. It should just be a type change

I would strongly prefer this as well.

SIPS/sip-29.md
```diff
@@ -35,85 +35,147 @@ in uppercase in this document are to be interpreted as described in [RFC
...
 
-#### Get Token Description
+Any Snap that wishes to provide asset information **MUST** implement the following API:
+
+#### Get Asset Description
+
+```typescript
+import { OnAssetDescriptionHandler } from "@metamask/snaps-sdk";
+
+export const onAssetDescription: OnAssetDescriptionHandler = async ({
```
FrederikBolding, line 70: I'm guessing this will be covered in another doc, but I'm curious how data is added to this list provided by the caller. How are assets discovered originally?

GuillaumeRx: This is covered by the kering API with the `listAccountAssets` method IIRC @danroc can provide more details about it

GuillaumeRx: Then the snap uses any API that provides asset metadata to create its token description

SIPS/sip-29.md
```diff
@@ -35,85 +35,147 @@ in uppercase in this document are to be interpreted as described in [RFC
...
+This SIP introduces a new permission named `endowment:assets`.
+This permission grants a Snap the ability to provide asset information to the client.
+
+This permission is specified as follows in `snap.manifest.json` files:
+
+```json
+{
+  "initialPermissions": {
+    "endowment:assets": {
+        "chains": [
```
danroc, line 49: Should we allow namespaces as well? And we should probably name it `scopes` to be aligned with SIP-26, CAIP-25/27.

danroc:
> > > I wasn't sure whether to use `asset` or `token` in the method names, as asset is more generic and could include NFTs, for instance, where the details payload would differ significantly from that of tokens.
> > 
> > 
> > I wonder if this API couldn't be versatile enough that we could add the support for NFT down the road rather than having multiple endpoints. It should just be a type change
> 
> I would strongly prefer this as well.

Me too :)

What we did in the `listAccountTransactions` was to create a tagged union (`{fungible: true; ... } | {fungible: false; ... }`), maybe we should apply the same pattern here, since fungible and non-fungible assets have very distinct properties.

SIPS/sip-29.md
GuillaumeRx: What's your thinking for allowing namespaces ? I would prefer this to be granular and to have to define reference supported rather than defining namespaces that would probably introduce calls for reference that are not supported

SIPS/sip-29.md
danroc: I think it makes more sense to have `{from, to}[]`

PR #164, opened 2025-03-19
SIP-31: Introduce `onClientRequest` entrypoint

danroc:
This proposal introduces a new entrypoint, `onClientRequest`, allowing Snaps to expose a handler that can only be called by clients.

While the request `origin` can currently be used to differentiate requests from clients and dapps, this new entrypoint increases separation and minimizes the risks associated with `origin` spoofing or request routing bugs.

### DIFF
--- a/SIPS/sip-31.md
+++ b/SIPS/sip-31.md
@@ -86,7 +86,7 @@ type OnClientRequestArguments = {
 The type for an `onClientRequest` handler function’s return value is:
 
 ```typescript
-type OnClientRequestReturnValue = Json;
+type OnClientRequestResponse = Json;
 ```
 
 #### Behavior
--- a/SIPS/sip-31.md
+++ b/SIPS/sip-31.md
@@ -46,21 +46,48 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 in uppercase in this document are to be interpreted as described in [RFC
 2119](https://www.ietf.org/rfc/rfc2119.txt).
 
-### `onClientRequest` Entrypoint
+### Snap Implementation
 
-Snaps MAY define an `onClientRequest` handler with the following signature:
+#### On Client Request
+
+This specification introduces to the Snap Platform a dedicated handler that
+Snaps MAY implement to process JSON-RPC requests originating exclusively from
+the client.
+
+Example:
 
 ```typescript
-type OnClientRequest = ({ request }: { request: JsonRpcRequest }) => Promise<Json>;
+import { OnClientRequestHandler } from "@metamask/snaps-sdk";
+
+export const onClientRequest: OnClientRequestHandler = async ({
+  request,
+}) => {
+  const result = /* Handle `request` */;
+  return result;
+};
 ```
 
-#### Parameters
+The type of the `onClientRequest` handler is:
 
-- `request: JsonRpcRequest` – A JSON-RPC request object.
+```typescript
+type OnClientRequestHandler = (
+  args: OnClientRequestArguments,
+) => Promise<OnClientRequestReturnValue>;
+```
 
-#### Returns
+The type for an `onClientRequest` handler function’s arguments is:
 
-A `Promise<Json>`, which resolves to any valid JSON structure.
+```typescript
+type OnClientRequestArguments = {
+  request: JsonRpcRequest;
+};
+```
+
+The type for an `onClientRequest` handler function’s return value is:
+
+```typescript
+type OnClientRequestReturnValue = Json;
+```
 
 #### Behavior
 
--- a/SIPS/sip-31.md
+++ b/SIPS/sip-31.md
@@ -9,35 +9,42 @@ created: 2025-03-19
 ## Abstract
 
 This proposal introduces a new entrypoint, `onClientRequest`, allowing Snaps to
-expose a handler that can only be called by MetaMask clients.
+expose a handler that can only be called by clients.
 
-While request `origin` is currently used to differentiate requests from
-MetaMask and dapps, this new entrypoint increases separation and minimizes the
-risks associated with `origin` spoofing or request routing bugs.
+While the request `origin` can currently be used to differentiate requests from
+clients and dapps, this new entrypoint increases separation and minimizes
+the risks associated with `origin` spoofing or request routing bugs.
 
 ## Motivation
 
 Currently, Snaps rely on the `origin` field of incoming requests to
-differentiate whether a request originates from MetaMask or an external dapp.
+differentiate whether a request originates from the client or an external dapp.
 However, this approach has potential risks:
 
 - **Origin Spoofing**: Bugs or vulnerabilities in request validation could
-  allow dApps to masquerade as MetaMask clients.
+  allow dapps to masquerade as clients.
 
-- **Unintended Exposure**: If a Snap processes requests without strict checks,
-  it may inadvertently handle dapp requests intended only for MetaMask.
+- **Unintended Exposure**: If a Snap processes requests without strict
+  validation, it may unintentionally expose to dapps methods that are meant
+  only for clients.
 
-- **Cleaner Separation**: Having a dedicated entrypoint for MetaMask requests
+- **Cleaner Separation**: Having a dedicated entrypoint for client requests
   enforces stricter request isolation at the API level.
 
 By introducing `onClientRequest`, we ensure that Snaps can define handlers
-exclusively accessible by MetaMask, reducing attack vectors and improving
-request security.
+exclusively accessible by clients, reducing attack vectors.
 
 ## Specification
 
-> This proposal introduces a new optional handler function, `onClientRequest`,
-> which a Snap can implement.
+This proposal introduces a new optional handler function, `onClientRequest`,
+which a Snap MAY implement.
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" written
+in uppercase in this document are to be interpreted as described in [RFC
+2119](https://www.ietf.org/rfc/rfc2119.txt).
 
 ### `onClientRequest` Entrypoint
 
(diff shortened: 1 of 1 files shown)


SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+which a Snap MAY implement.
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" written
+in uppercase in this document are to be interpreted as described in [RFC
+2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+### `onClientRequest` Entrypoint
```
danroc, line 49: Should we also define a new endowment?

Montoya: We don't need it if this is restricted to pre-installed Snaps right?

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+title: Client-only RPC Entrypoint
+status: Draft
+author: Daniel Rocha (@danroc)
+created: 2025-03-19
+---
+
+## Abstract
+
+This proposal introduces a new entrypoint, `onClientRequest`, allowing Snaps to
+expose a handler that can only be called by clients.
```
Montoya, line 12: by _the client_. (singular)

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+
+While the request `origin` can currently be used to differentiate requests from
+clients and dapps, this new entrypoint increases separation and minimizes
+the risks associated with `origin` spoofing or request routing bugs.
+
+## Motivation
+
+Currently, Snaps rely on the `origin` field of incoming requests to
+differentiate whether a request originates from the client or an external dapp.
```
Montoya, line 21: whether a request originates from the client, _another Snap,_ or an external dapp.

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+However, this approach has potential risks:
+
+- **Origin Spoofing**: Bugs or vulnerabilities in request validation could
+  allow dapps to masquerade as clients.
```
Montoya, line 25: as _the client_.

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+
+- **Unintended Exposure**: If a Snap processes requests without strict
+  validation, it may unintentionally expose to dapps methods that are meant
```
Montoya, line 28: expose methods _to dapps or other Snaps_ that are meant only for _the client_.

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+  only for clients.
+
+- **Cleaner Separation**: Having a dedicated entrypoint for client requests
+  enforces stricter request isolation at the API level.
+
+By introducing `onClientRequest`, we ensure that Snaps can define handlers
+exclusively accessible by clients, reducing attack vectors.
```
Montoya, line 35: by _the client_, reducing _potential_ attack vectors.

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Specification
+
+This proposal introduces a new optional handler function, `onClientRequest`,
```
Montoya, line 39: optional could be all-caps OPTIONAL

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+
+- `request: JsonRpcRequest` – A JSON-RPC request object.
+
+#### Returns
+
+A `Promise<JsonRpcResponse>`, which resolves to a JSON-RPC response object.
+
+#### Behavior
+
+- Clients MUST ensure that they are the origin of a request before invoking
```
Montoya, line 67: Can also make this singular. 

- The client MUST ensure that it is the origin of the request before invoking

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+  `onClientRequest`.
+
+- Requests from other origins MUST be rejected.
+
+- If a Snap does not implement `onClientRequest`, the client MUST throw an
+  exception for any request, regardless of the origin.
```
Montoya, line 73: is this correct? any request? the Snap can still on `onRPCRequest` and other request interfaces right?

Montoya (changes requested):

SIPS/sip-31.md
danroc: This sentence isn't clear, I meant requests to the `onClientRequest` entrypoint. I just updated it.

SIPS/sip-31.md
danroc: Yes, but should we limit this entrypoint to pre-installed Snaps?

Montoya: Well I'm only speculating so maybe this is not relevant, but my thinking is: 

1. We only plan to implement `onClientRequest` in flows for pre-installed Snaps. We don't have any plans to do this with any 3rd party Snaps yet. 
2. The biggest risk, which may or may not be relevant, is if a 3rd party Snap spoofs itself as a pre-installed Snap (some 3rd party Snap pretends to be the ENS Resolver Snap, or MMI Snap, for example), and thus gains malicious access. 

So this is why I think we either need to introduce an endowment or make it only for pre-installed Snaps. I don't think we can skip both options. But would like someone else to validate or correct my thinking here...

Montoya (approved):
Generally approve but I think [this conversation](https://github.com/MetaMask/SIPs/pull/164/files#r2026398759) should be resolved.

SIPS/sip-31.md
FrederikBolding: Is there a use-case for `onClientRequest` beyond preinstalled Snaps? 

My assumption was that we wanted `onClientRequest` for non-standard JSON-RPC requests sent to Snaps in one way or another affiliated with MetaMask. Generally standardized JSON-RPC for everyone (third parties and preinstalled Snaps) should have be separate standardized handlers (e.g. `onProtocolRequest`).

Montoya: OK so sounds like we are on board to make this only for pre-installed Snaps. Sounds good to me.

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" written
+in uppercase in this document are to be interpreted as described in [RFC
+2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+### `onClientRequest` Entrypoint
+
+Snaps MAY define an `onClientRequest` handler with the following signature:
+
+```typescript
+type OnClientRequest = (request: JsonRpcRequest) => Promise<JsonRpcResponse>;
```
FrederikBolding, line 54:
```suggestion
type OnClientRequest = (request: JsonRpcRequest) => Promise<Json>;
```

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+#### Parameters
+
+- `request: JsonRpcRequest` – A JSON-RPC request object.
+
+#### Returns
+
+A `Promise<JsonRpcResponse>`, which resolves to a JSON-RPC response object.
```
FrederikBolding, line 63:
```suggestion
A `Promise<Json>`, which resolves to any valid JSON structure.
```

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+which a Snap MAY implement.
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" written
+in uppercase in this document are to be interpreted as described in [RFC
+2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+### `onClientRequest` Entrypoint
```
FrederikBolding, line 49: Can we use the same formatting for this section as https://metamask.github.io/SIPs/SIPS/sip-29 ?

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+
+- `request: JsonRpcRequest` – A JSON-RPC request object.
+
+#### Returns
+
+A `Promise<JsonRpcResponse>`, which resolves to a JSON-RPC response object.
+
+#### Behavior
+
+- The client MUST ensure that they are the origin of a request before invoking
```
FrederikBolding, line 67: Hmm, wondering if we wanna be more explicit about how to do this.

In practice we would probably check `origin === 'metamask'`, but that is of course an implementation detail.

SIPS/sip-31.md
```diff
@@ -0,0 +1,77 @@
...
+the risks associated with `origin` spoofing or request routing bugs.
+
+## Motivation
+
+Currently, Snaps rely on the `origin` field of incoming requests to
+differentiate whether a request originates from the client, another
+Snap, or an external dapp. However, this approach has potential risks:
+
+- **Origin Spoofing**: Bugs or vulnerabilities in request validation could
+  allow dapps to masquerade as the client.
```
FrederikBolding, line 25: I guess this somewhat depends on what kind of origin spoofing is possible. I would assume that the most likely case is not that a dapp can masquerade as the client, but instead as another dapp.

SIPS/sip-31.md
```diff
@@ -0,0 +1,104 @@
...
+  return result;
+};
+```
+
+The type of the `onClientRequest` handler is:
+
+```typescript
+type OnClientRequestHandler = (
+  args: OnClientRequestArguments,
+) => Promise<OnClientRequestReturnValue>;
```
FrederikBolding, line 75:
```suggestion
) => Promise<OnClientRequestResponse>;
```

SIPS/sip-31.md
Montoya: Also realizing it should be "The client MUST ensure that **it is** the origin..."

FrederikBolding:
```suggestion
- The client MUST ensure that it is the origin of a request before invoking
```

Montoya: Maybe be explicit by giving an example with the MetaMask client

FrederikBolding (approved):

Montoya (approved):

PR #165, opened 2025-03-20
Introduce `snap_trackEvent` method

danroc:
### Description

This Snap Improvement Proposal (SIP) introduces a new method, `snap_trackEvent`, allowing pre-installed Snaps to submit tracking events through the client. The client will determine how to handle these events.

### DIFF
--- a/SIPS/sip-32.md
+++ b/SIPS/sip-32.md
@@ -27,12 +27,11 @@ analytics or tracking events. This proposal aims to:
 
 ## Specification
 
-> The `snap_trackEvent` method allows pre-installed Snaps to submit structured
-> tracking events to the client. It is the client’s responsibility to process
-> these events accordingly.
->
-> It's the client's responsibility to ensure that only pre-installed Snaps can
-> call this method.
+The `snap_trackEvent` method allows pre-installed Snaps to submit structured
+tracking events to the client. It is the client’s responsibility to process
+these events accordingly.
+
+The client MUST ensure that only pre-installed Snaps can call this method.
 
 ### `snap_trackEvent`
 
--- a/SIPS/sip-32.md
+++ b/SIPS/sip-32.md
@@ -84,11 +84,6 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 in uppercase in this document are to be interpreted as described in [RFC
 2119](https://www.ietf.org/rfc/rfc2119.txt)
 
-## Backwards compatibility
-
-Any SIPs that break backwards compatibility MUST include a section describing
-those incompatibilities and their severity. The SIP SHOULD describe how the
-author plans on proposes to deal with such these incompatibilities.
 
 ## Copyright
 
--- a/SIPS/sip-32.md
+++ b/SIPS/sip-32.md
@@ -41,9 +41,8 @@ The `snap_trackEvent` method allows pre-installed Snaps to submit structured
 tracking events to the client. It is the client’s responsibility to process
 these events accordingly.
 
-The client MUST ensure that only pre-installed Snaps can call this method.
-
-The `snap_trackEvent` method is defined as follows:
+The `snap_trackEvent` method is endowed on any Snap that the client considers
+pre-installed.
 
 #### Parameters
 
--- a/SIPS/sip-32.md
+++ b/SIPS/sip-32.md
@@ -20,7 +20,8 @@ event tracking while maintaining user privacy.
 Currently, there is no standardized way for pre-installed Snaps to submit
 analytics or tracking events. This proposal aims to:
 
-- Enable pre-installed Snaps to reuse MetaMask’s event tracking infrastructure.
+- Enable pre-installed Snaps to leverage the client's event tracking
+  infrastructure.
 
 - Ensure that event handling is controlled at the client level and conforms to
   security and privacy controls enforced by the client.
--- a/SIPS/sip-32.md
+++ b/SIPS/sip-32.md
@@ -12,8 +12,8 @@ This Snap Improvement Proposal (SIP) introduces a new method,
 `snap_trackEvent`, allowing pre-installed Snaps to submit tracking events
 through the client. The client will determine how to handle these events.
 
-This feature enables Snaps to utilize the existing MetaMask client
-infrastructure for event tracking while maintaining user privacy.
+This feature enables Snaps to utilize the existing client infrastructure for
+event tracking while maintaining user privacy.
 
 ## Motivation
 
@@ -22,8 +22,8 @@ analytics or tracking events. This proposal aims to:
 
 - Enable pre-installed Snaps to reuse MetaMask’s event tracking infrastructure.
 
-- Allow the MetaMask client to control event handling, ensuring security and
-  compliance with privacy policies.
+- Ensure that event handling is controlled at the client level and conforms to
+  security and privacy controls enforced by the client.
 
 ## Specification
 
--- a/SIPS/sip-32.md
+++ b/SIPS/sip-32.md
@@ -1,0 +1,95 @@
+---
+sip: 32
+title: Add `snap_trackEvent` method for pre-installed Snaps
+status: Draft
+author: Daniel Rocha (@danroc)
+created: 2025-03-20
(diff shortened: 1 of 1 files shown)


SIPS/sip-32.md
```diff
@@ -0,0 +1,95 @@
...
+  compliance with privacy policies.
+
+## Specification
+
+> The `snap_trackEvent` method allows pre-installed Snaps to submit structured
+> tracking events to the client. It is the client’s responsibility to process
+> these events accordingly.
+>
+> It's the client's responsibility to ensure that only pre-installed Snaps can
+> call this method.
```
FrederikBolding, line 35:
```suggestion
The `snap_trackEvent` method allows pre-installed Snaps to submit structured
tracking events to the client. It is the client’s responsibility to process
these events accordingly.

The client MUST ensure that only pre-installed Snaps can call this method.
```

SIPS/sip-32.md
```diff
@@ -0,0 +1,95 @@
...
+
+  - `name: string` - The name of the event to track.
+
+  - `properties: Record<string, Json>` - (**Optional**) Custom values to track.
+    Keys in this object must be `snake_case`.
+
+  - `sensitiveProperties: Record<string, Json>` - (**Optional**) Sensitive
+    values to track. These properties will be sent in an additional event that
+    excludes the user's `metaMetricsId`. Keys in this object must be in
+    `snake_case`.
```
FrederikBolding, line 54: Should the format of these actually be validated?

Please use the MUST/SHOULD keywords

SIPS/sip-32.md
```diff
@@ -0,0 +1,95 @@
...
+### `snap_trackEvent`
+
+The `snap_trackEvent` method is defined as follows:
+
+#### Parameters
+
+- `event` - Event object to be tracked.
+  - `category: string` - The category to associate the event.
+
+  - `name: string` - The name of the event to track.
```
FrederikBolding, line 46: Do we need to uphold any rules for the format for these?

SIPS/sip-32.md
```diff
@@ -0,0 +1,95 @@
...
+  },
+});
+```
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" written
+in uppercase in this document are to be interpreted as described in [RFC
+2119](https://www.ietf.org/rfc/rfc2119.txt)
```
FrederikBolding, line 85: Please move this to the top of the spec

SIPS/sip-32.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Backwards compatibility
+
+Any SIPs that break backwards compatibility MUST include a section describing
+those incompatibilities and their severity. The SIP SHOULD describe how the
+author plans on proposes to deal with such these incompatibilities.
```
FrederikBolding, line 91:
```suggestion
```

SIPS/sip-32.md
danroc: I don't think so, at least I didn't find any recommendation about it. I'll try to check with the data team.

danroc: I got this answer from data:

> The constraints what we impose semantically in order to preserve the continuity of the analytics; same name and same format as the event fired by extension, newly installed snaps, etc.

It seems that we cannot enforce a name format here.

SIPS/sip-32.md
```diff
@@ -0,0 +1,89 @@
...
+The client MUST ensure that only pre-installed Snaps can call this method.
+
+The `snap_trackEvent` method is defined as follows:
+
+#### Parameters
+
+- `event` - Event object to be tracked.
+  - `category: string` - The category to associate the event.
+
+  - `name: string` - The name of the event to track.
```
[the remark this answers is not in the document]
FrederikBolding: Hmm, wouldn't `event.event` be easier so we don't require mapping?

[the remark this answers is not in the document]
danroc: Probably yes, I just didn't want to leak this into the API, but I don't have a strong opinion on this.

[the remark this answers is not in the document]
Montoya: I'm fine with `event.event`

SIPS/sip-32.md
```diff
@@ -0,0 +1,89 @@
...
+
+This feature enables Snaps to utilize the existing MetaMask client
+infrastructure for event tracking while maintaining user privacy.
+
+## Motivation
+
+Currently, there is no standardized way for pre-installed Snaps to submit
+analytics or tracking events. This proposal aims to:
+
+- Enable pre-installed Snaps to reuse MetaMask’s event tracking infrastructure.
```
Montoya, line 23: Enable pre-installed Snaps to _leverage_ MetaMask's event tracking infrastructure.

SIPS/sip-32.md
```diff
@@ -0,0 +1,89 @@
...
+created: 2025-03-20
+---
+
+## Abstract
+
+This Snap Improvement Proposal (SIP) introduces a new method,
+`snap_trackEvent`, allowing pre-installed Snaps to submit tracking events
+through the client. The client will determine how to handle these events.
+
+This feature enables Snaps to utilize the existing MetaMask client
```
Montoya, line 15: the existing client infrastructure (drop MetaMask)

SIPS/sip-32.md
```diff
@@ -0,0 +1,89 @@
...
+in uppercase in this document are to be interpreted as described in [RFC
+2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### `snap_trackEvent`
+
+The `snap_trackEvent` method allows pre-installed Snaps to submit structured
+tracking events to the client. It is the client’s responsibility to process
+these events accordingly.
+
+The client MUST ensure that only pre-installed Snaps can call this method.
```
Montoya, line 43: Is there some endowment that guarantees this? Is the endowment only available to pre-installed Snaps? I think we need to document how this is enforced.

Montoya (changes requested):
Left multiple comments.

SIPS/sip-32.md
```diff
@@ -0,0 +1,89 @@
...
+#### Parameters
+
+- `event` - Event object to be tracked.
+  - `category: string` - The category to associate the event.
+
+  - `name: string` - The name of the event to track.
+
+  - `properties: Record<string, Json>` - (**Optional**) Custom values to track.
+    The client MUST enforce that all keys in this object are in the
+    `snake_case` format.
```
Mrtenz, line 56: Is this actually a requirement for our analytics, or simply a convention?

SIPS/sip-32.md
danroc: I think it's an implementation detail _how_ the client enforces it.

FrederikBolding: Not necessarily, it is important to include in the spec if a new permission is required.

That being said, I think we could make this available to all preinstalled Snaps without a permission/endowment.

WDYT @Mrtenz?

Mrtenz: > That being said, I think we could make this available to all preinstalled Snaps without a permission/endowment.

I agree, I don't see a reason to make this permissioned.

FrederikBolding: For the specs sake you can just add something like
```
`snap_trackEvent` is endowed on any Snap that the client considers preinstalled.
```

Similar to https://metamask.github.io/SIPs/SIPS/sip-30

SIPS/sip-32.md
danroc: The data team might want to enforce it in the future, but currently it's only a convention

FrederikBolding (approved):

SIPS/sip-32.md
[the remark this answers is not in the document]
danroc: Reopening this. @Vahid-flipside shared an internal document that uses `name` instead, so it looks like the extension should be updated to use `name` instead of `event`.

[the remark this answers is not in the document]
Montoya: OK event.name is fine

PR #167, opened 2025-04-02
fix issues in SIP-29

GuillaumeRx:
This PR fixes some issues with SIP-29:
- Inconsistencies with the historical price handler - `onAssetHistoricalPrice`
- Fixes the `onAssetHistoricalPrice` handlers return type
- ERC-20 standard makes the `name` and `symbol` optional, this aligns the `FungibleToken` type with it.

### DIFF
--- a/SIPS/sip-29.md
+++ b/SIPS/sip-29.md
@@ -4,7 +4,7 @@ title: Snap Assets API
 status: Draft
 author: Daniel Rocha (@danroc), Guillaume Roux (@GuillaumeRx)
 created: 2024-12-05
-updated: 2025-01-25
+updated: 2025-04-02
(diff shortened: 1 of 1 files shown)


SIPS/sip-29.md
```diff
@@ -225,10 +229,10 @@ type FungibleAssetUnit = {
 // Fungible asset metadata.
 type FungibleAssetMetadata = {
   // Human-friendly name of the asset.
-  name: string;
+  name?: string;
```
FrederikBolding, line 232: Does the asset controllers and UI support this? If not, we may need to keep this as is

FrederikBolding (approved):

PR #171, opened 2025-04-08
SIP-32: Remove unnecessary field of event payload

danroc:
The PR remove `category` from the `event` object. @worldlyjohn pointed out that it's legacy and shouldn't be used anymore.

### DIFF
--- a/SIPS/sip-32.md
+++ b/SIPS/sip-32.md
@@ -47,18 +47,11 @@ pre-installed.
 #### Parameters
 
 - `event` - Event object to be tracked.
-  - `category: string` - The category to associate the event.
-
-  - `event: string` - The name of the event to track.
+  - `name: string` - The name of the event to track.
 
   - `properties: Record<string, Json>` - (**Optional**) Custom values to track.
     The client MUST enforce that all keys in this object are in the
     `snake_case` format.
-
-  - `sensitiveProperties: Record<string, Json>` - (**Optional**) Sensitive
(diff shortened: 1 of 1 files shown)


SIPS/sip-32.md
```diff
@@ -47,19 +47,12 @@ pre-installed.
 #### Parameters
 
 - `event` - Event object to be tracked.
-  - `category: string` - The category to associate the event.
-
-  - `event: string` - The name of the event to track.
+  - `name: string` - The name of the event to track.
```
FrederikBolding, line 50: Why are we changing this name again?

SIPS/sip-32.md
```diff
@@ -47,19 +47,12 @@ pre-installed.
...
 
   - `properties: Record<string, Json>` - (**Optional**) Custom values to track.
     The client MUST enforce that all keys in this object are in the
     `snake_case` format.
 
-  - `sensitiveProperties: Record<string, Json>` - (**Optional**) Sensitive
```
FrederikBolding, line 58: As far as I can tell we cannot remove this

SIPS/sip-32.md
danroc: Reverting the change for now

SIPS/sip-32.md
danroc: Reverting this until a decision is made

FrederikBolding (approved):

PR #172, opened 2025-04-09
SIP-34: Introduce `endowment:keyring.version`

ccharly:
## Description

Introducing a new `version` field in the `endowment:keyring` configuration object, to allow conditional logic when introducing new breaking change to the keyring API.

### DIFF
--- a/SIPS/sip-34.md
+++ b/SIPS/sip-34.md
@@ -45,7 +45,7 @@ This field will be a `number` and will defaults to `1` if not provided to mainta
 
 On the keyring API implementation side, we will modify our API methods to conditionally behave based on the `version`:
 - When `version` is set to `1`, the API will behave as it currently does.
-- When `version` is set to `2`, the API will introduce new incompatible behaviors/features with verison `1`.
+- When `version` is set to `2`, the API will introduce new incompatible behaviors/features with version `1`.
 - Future versions will allow additional breaking features, with conditional logic to ensure backward compatibility.
 
 This new implementation MUST ensure backward compatibility with older Snaps.
--- a/SIPS/sip-34.md
+++ b/SIPS/sip-34.md
@@ -1,0 +1,87 @@
+---
+sip: 34
+title: endowment:keyring.version
+status: Draft
+discussions-to: https://github.com/MetaMask/SIPs/discussions/172
+author: Charly Chevalier (@ccharly), Daniel Rocha (@danroc)
+created: 2025-04-07
+---
+
+## Abstract
+
+This proposal introduces a new field, `version`, within the existing configuration object used by the `endowment:keyring`.
+This field will allow the versioning of the keyring API, providing the flexibility to define the current API version being used.
+By declaring the version in the configuration, we can implement conditional behavior in the API based on the selected version.
+This will allow us to modify or add new functionality in the API without breaking backward compatibility.
+
+## Motivation
+
+The current `endowment:keyring` configuration does not support versioning, which limits our ability to introduce breaking changes or new functionality that may require changes in behavior.
+Adding the `version` field allows us to version the API and define distinct behaviors based on the declared version.
+
+This is useful as we want to add new fields and features that should only apply to specific versions of the API, ensuring that existing implementations relying on previous versions continue to work without modification.
+
+Our current request "parsing" (within the keyring API) is strict and does not allow extra-fields. Meaning that current EVM Snaps might throw an error if we add new fields to some of our requests (even optional fields).
+
+We plan to modify the `submitRequest(request)` method by adding a new `origin` field to the `request` parameter, but only for version `2` and higher.
+We won't forward this `origin` field for requests made with version `1` to ensure backward compatibility.
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": "7c507ff0-365f-4de0-8cd5-eb83c30ebda4",
+  "method": "keyring_submitRequest",
+  "params": {
+    "id": "c555de37-cf4b-4ff2-8273-39db7fb58f1c",
+    "scope": "eip155:1",
+    "account": "4abdd17e-8b0f-4d06-a017-947a64823b3d",
+    "request": {
+      "method": "eth_method",
+      "params": [1, 2, 3]
+    },
+    "origin": "someOrigin"
+  }
+}
+```
+
+This system will allow us to introduce future versions in a flexible way, limiting the need for breaking changes when new features are added.
+
+## Specification
+
+### Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+### Proposed implementation
+
+Add an OPTIONAL `version` field to the `endowment:keyring` configuration object.
+This field will be a `number` and will defaults to `1` if not provided to maintain compatibility with existing implementations.
+
+   ```json
+   {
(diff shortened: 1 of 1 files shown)


SIPS/sip-34.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Proposed implementation
+
+Add an OPTIONAL `version` field to the `endowment:keyring` configuration object.
+This field will be a `number` and will defaults to `1` if not provided to maintain compatibility with existing implementations.
+
+   ```json
+   {
+     "endowment:keyring": {
+       "version": 2,
```
ccharly, line 63: **Question 1**: I initially wanted to use `keyringApiVersion`, but that does sounds a bit "too much" (since we're already in a "keyring context", but maybe `apiVersion` could work?

- - -

**Question 2**: I thought using `number` would make it easier to write runtime-logic like:

```ts
if (manifest['endowment:keyring'].version >= api.version) {
  // Then we can use breaking features from version `api.version`
}
```

We could also go with a "version-`string`", like `'v1'`, `'v2'`, etc... But they might not be really practical to use (unless we are converting them to actual `number` under the hood)

Mrtenz: I don't think semver makes much sense here. The main purpose of this version is to introduce breaking changes. Non-breaking changes don't need a version increment.

SIPS/sip-34.md
```diff
@@ -0,0 +1,87 @@
...
+title: endowment:keyring.version
+status: Draft
+discussions-to: https://github.com/MetaMask/SIPs/discussions/172
+author: Charly Chevalier (@ccharly), Daniel Rocha (@danroc)
+created: 2025-04-07
+---
+
+## Abstract
+
+This proposal introduces a new field, `version`, within the existing configuration object used by the `endowment:keyring`.
```
Mrtenz, line 12: I wonder if we should make this more generic, and instead describe a way for any endowment, or permission in general, to specify a version.

SIPS/sip-34.md
GuillaumeRx: Makes sense ! I was wondering if we would not ever do something like we did for `snaps-sdk` in the manifest for the keyring-api but I guess this is unrelated.

Mrtenz: Yeah, we could potentially leverage the `platformVersion` as well instead of having a version for each permission separately.

ccharly: Given that the `keyring-api` is not "related" to the platform itself, IDK if the `platformVersion` could be used in our use-case. 🤔

But I agree with @Mrtenz here, this version serves a different purpose, and having a single `number` should work well and be enough.

SIPS/sip-34.md
ccharly: I'm definitely ok with that, as long as we have a way to use that version, I'm fine with this.

This would probably allow breaking changes within the endowments themselves yes!

SIPS/sip-34.md
```diff
@@ -0,0 +1,87 @@
...
+## Motivation
+
+The current `endowment:keyring` configuration does not support versioning, which limits our ability to introduce breaking changes or new functionality that may require changes in behavior.
+Adding the `version` field allows us to version the API and define distinct behaviors based on the declared version.
+
+This is useful as we want to add new fields and features that should only apply to specific versions of the API, ensuring that existing implementations relying on previous versions continue to work without modification.
+
+Our current request "parsing" (within the keyring API) is strict and does not allow extra-fields. Meaning that current EVM Snaps might throw an error if we add new fields to some of our requests (even optional fields).
+
+We plan to modify the `submitRequest(request)` method by adding a new `origin` field to the `request` parameter, but only for version `2` and higher.
```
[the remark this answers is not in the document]
ccharly: Sounds like a good plan to me yes! First, the versioning, second, the actual change introduced with that version.

I'll get rid of this detailed example then, thanks!

SIPS/sip-34.md
```diff
@@ -0,0 +1,87 @@
...
+We won't forward this `origin` field for requests made with version `1` to ensure backward compatibility.
```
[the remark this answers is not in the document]
ccharly: Yes.. So you are right, new field shouldn't be considered breaking, however, in our current "Keyring API SDK", we are exposing some functions to dispatch incoming requests into the right "handler" using `handleKeyringRequest`:
- https://github.com/MetaMask/accounts/blob/main/packages/keyring-snap-sdk/src/rpc-handler.ts#L190-L223
- https://github.com/MetaMask/accounts/blob/main/packages/keyring-snap-sdk/src/rpc-handler.ts#L163-L166

And right now, those functions are "strict" and will `superstruct.assert`, so if we have any extra fields that are not part of those structs, `assert` will throw.

That's probably a bad design choice from us here 😅 but if we start to add new field, older Snaps that use those functions will end up with an error.

An improvement to this could be to allow additional fields and mention to the Snap developers to allow extra-fields as well (in case they are doing their own dispatching and do not rely on our SDK).

So that's why we consider this a breaking change in this context.

SIPS/sip-34.md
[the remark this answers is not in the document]
ccharly: [922ce85](https://github.com/MetaMask/SIPs/pull/172/commits/922ce85eebbb25304bdfcbaa3448cdb4116c886d)

SIPS/sip-34.md
Mrtenz: > Given that the `keyring-api` is not "related" to the platform itself, IDK if the `platformVersion` could be used in our use-case. 🤔

It's somewhat related, since the Snaps keyring API is built on the Snaps platform 😅 I wonder if instead of adding a version for specific permissions, we should check if the platform version is newer than a certain version instead.

ccharly: @danroc any thoughts on this?

my only worry is, if we make a new keyring API breaking change (like for the old `submitRequest` example with the `origin`), we would make a new major version of the keyring API, but the platform version won't be bumped at the same time, since the keyring API is not really packaged/bundled/part of the platform itself? (the `endowment:keyring` is part of it though)

NOTE: I have no idea how/when we bump the platform version 😅

danroc: I don't have a strong opinion on this. However, we will probably need to make another breaking change at some point later this year (we want to split the Keyring API into smaller interfaces to cover different use cases).

In my opinion, it's simpler to handle a single platform version. It would prevent us from having to check for different combinations of endowment **X** versus endowment **Y**, but it also means we might need to bump the global version more often. That said, this shouldn't happen too frequently.

SIPS/sip-34.md
```diff
@@ -0,0 +1,58 @@
...
+       "allowedOrigins": [
+         ...
+       ]
+     }
+   }
+   ```
+
+On the keyring API implementation side, we will modify our API methods to conditionally behave based on the `version`:
+- When `version` is set to `1`, the API will behave as it currently does.
+- When `version` is set to `2`, the API will introduce new incompatible behaviors/features with verison `1`.
```
shane-t, line 48:
```suggestion
- When `version` is set to `2`, the API will introduce new incompatible behaviors/features with version `1`.
```

ccharly:
After some internal discussions with @FrederikBolding @Mrtenz @danroc and I. It has been decided to use the next major version of the Snap platform as the "breaking change point" to introduce the new `KeyringRequest.origin`.

We also talked about future ways of integrating and coupling the keyring API within the platform itself. Having this API being integrated within the platform could help when we need to update the "transport layer" itself.

For the moment, we rely on the new action `:isMinimumPlatformVersion` to prevent sending the `origin` field if a Snap does not declare a `platformVersion >= 7.0.0`.

See:
- https://github.com/MetaMask/snaps/blob/main/packages/snaps-sdk/CHANGELOG.md#700
- https://github.com/MetaMask/accounts/blob/main/packages/keyring-api/CHANGELOG.md#1800

We don't need this extra `endowment:keyring.version` anymore, closing this.

PR #174, opened 2025-05-05
SIP-29: Add support for non-fungible assets

GuillaumeRx:


SIPS/sip-29.md
```diff
@@ -59,13 +59,13 @@ This permission is specified as follows in `snap.manifest.json` files:
...
 
-Two methods are defined in the Snap Assets API:
+Four methods are defined in the Snap Assets API:
 
 Any Snap that wishes to provide asset information MUST implement the following API:
 
 #### Get Assets Metadata
 
-`Caip19AssetType` - A string that represents an asset using the [CAIP-19][caip-19] standard.
+`Caip19AssetTypeOrId` - A string that represents an asset using the [CAIP-19][caip-19] standard.
```
FrederikBolding, line 68: I don't follow the name change here?

GuillaumeRx: On some chains NFTs uses CAIP-19 asset IDs (like EVM) and on some others it's CAIP-19 asset types (like Solana)

SIPS/sip-29.md
```diff
@@ -243,9 +267,126 @@ type FungibleAssetMetadata = {
...
+
+  // The last sale of one asset in the collection.
+  lastSale?: {
+    // The asset that was sold.
+    asset: Caip19AssetTypeOrId;
+    // The price at which is was sold represented as a decimal number in a string.
+    amount: string;
+  };
+
+  // The top bid on the asset.
```
sahar-fehri, line 364: I like that all attributes are optional in `NonFungibleAssetMarketData` 🙏 
In case we have new marketData coming from NFT-API, will we have to update the snap first?

GuillaumeRx: From the types you sent me it seems like all of these can be optional :)

If we add new data from the NFT-API we will have to update the platform if it's not yet specified here and then update the snap yes !

SIPS/sip-29.md
```diff
@@ -243,9 +267,126 @@ type FungibleAssetMetadata = {
...
+  name?: string;
+
+  // Ticker symbol of the asset.
+  symbol?: string;
+
+  // Base64 data URI or URL representation of the asset image.
+  imageUrl: string;
+
+  // The description of the asset.
+  description?: string;
```
sahar-fehri, line 337: I believe we also want to add
```
isSpam?: boolean
acquired_at?: string
```

We use blockaid as a safety net for spam filtering, would be used on controller side to filter out spam nfts;

GuillaumeRx: for `isSpam` I would say the snap can handle that filtering, this seems like a very "only us" related field

Agree that `acquired_at` needs to be included !

SIPS/sip-29.md
```diff
@@ -117,15 +117,13 @@ type Conversion = {
 
 type OnAssetsConversionArguments = {
   conversions: Conversion[];
-  includeMarketData?: boolean;
```
sahar-fehri, line 120: Do we no longer need to pass this arg from the controller? is the snap going to automatically return market data for assets now?

GuillaumeRx: No we moved the market data to a new handler, we will have to call this new handler in the controller

GuillaumeRx: https://github.com/MetaMask/SIPs/blob/78ce26983b202dfb3828f4948972b62c07a66587/SIPS/sip-29.md?#L149-L187

SIPS/sip-29.md
FrederikBolding: Wait what? Why doesn't NFTs use caip19 asset IDs?

GuillaumeRx: No, it depends on the chain, solana uses asset types :sweat_smile:

GuillaumeRx: https://namespaces.chainagnostic.org/solana/caip19#examples

SIPS/sip-29.md
```diff
@@ -208,11 +230,13 @@ type OnAssetHistoricalPriceResponse = {
 };
 ```
 
-## Appendix I: Fungible Asset Metadata
+## Appendix I: Fungible Asset Metadata & Market Data
```
FrederikBolding, line 233:
```suggestion
## Appendix I: Asset Metadata & Market Data
```

SIPS/sip-29.md
```diff
@@ -208,11 +230,13 @@ type OnAssetHistoricalPriceResponse = {
...
 
-The following asset metadata fields for a fungible asset are defined.
+The following asset metadata and market data fields for a fungible and non-fungible asset are defined.
```
FrederikBolding, line 235:
```suggestion
The following asset metadata and market data fields for a fungible asset are defined.
```

SIPS/sip-29.md
```diff
@@ -243,9 +267,130 @@ type FungibleAssetMetadata = {
...
+  symbol: string;
+
+  // The number of tokens in the collection.
+  tokenCount: string;
+
+  // The creator address of the asset.
+  creator?: Caip19AssetType;
+
+  // Base64 data URI or URL representation of the asset icon.
+  imageUrl: string;
```
FrederikBolding, line 323: Should this be optional?

SIPS/sip-29.md
```diff
@@ -243,9 +267,130 @@ type FungibleAssetMetadata = {
...
+  name: string;
+
+  // Ticker symbol of the asset collection.
+  symbol: string;
+
+  // The number of tokens in the collection.
+  tokenCount: string;
+
+  // The creator address of the asset.
+  creator?: Caip19AssetType;
```
FrederikBolding, line 320: This type seems like a typo?

SIPS/sip-29.md
```diff
@@ -243,9 +267,130 @@ type FungibleAssetMetadata = {
...
+
+  // The description of the asset.
+  description?: string;
+
+  // Represents a non-fungible asset
+  fungible: false;
+
+  // The time at which the asset was acquired.
+  // The time is represented as a UNIX timestamp. 
+  acquired_at?: number;
```
FrederikBolding, line 344:
```suggestion
  acquiredAt?: number;
```

SIPS/sip-29.md
```diff
@@ -243,9 +267,130 @@ type FungibleAssetMetadata = {
...
+
+  // Attributes of the non-fungible asset.
+  attributes?: Record<string, string | number>;
+
+  // The collection of the asset.
+  collection: NonFungibleAssetCollection;
```
FrederikBolding, line 350: Should this be optional?

SIPS/sip-29.md
```diff
@@ -136,35 +134,59 @@ type AssetConversionRate = {
...
+  const marketData = /* Get market data for given `assets` */;
+  return { marketData };
+};
+```
+
+The type for an `onAssetsMarketData` handler function’s arguments is:
+
+```typescript
+type AssetPair = {
+  from: Caip19AssetTypeOrId;
```
danroc, line 166: IMO, we shouldn’t reuse `AssetPair` in the arguments of `onAssetMarketData`. The terms from and to don’t make much sense in this context since there is no conversion involved.

Something like `{ asset, unit }` would probably be more explicit.

SIPS/sip-29.md
```diff
@@ -243,13 +266,138 @@ type FungibleAssetMetadata = {
...
+  tokenCount: string;
+
+  // The creator address of the asset.
+  creator?: Caip10Address;
+
+  // Base64 data URI or URL representation of the asset icon.
+  imageUrl?: string;
+};
+
+type NonFungibleAssetMetadata = {
```
danroc, line 328: This type should probably have a `fungible: false` flag too.

GuillaumeRx: https://github.com/MetaMask/SIPs/blob/a30f9a0234dd67628a39fe997260035efd440984/SIPS/sip-29.md?plain=1#L345

GuillaumeRx: We have both :thinking:

SIPS/sip-29.md
```diff
@@ -243,13 +269,138 @@ type FungibleAssetMetadata = {
...
+  // Human-friendly name of the asset.
+  name?: string;
```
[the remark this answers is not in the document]
GuillaumeRx: What does `balance` means on a non fungible asset ?

`address` is found in the CAIP asset id/type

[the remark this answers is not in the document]
sahar-fehri: Good question; i still haven't seen the nft-api return a balance >1 in the response;
but i think in solana you can have multiple units of the same asset in which case the balance returned should be more than 1

SIPS/sip-29.md
```diff
@@ -243,13 +269,138 @@ type FungibleAssetMetadata = {
...
+```typescript
+type NonFungibleAssetCollection = {
+  // Human-friendly name of the asset collection.
+  name: string;
+
+  // Ticker symbol of the asset collection.
+  symbol: string;
+
+  // The number of tokens in the collection.
+  tokenCount: string;
```
sahar-fehri, line 322: wonder if this should be string | null; 
i know we are still experimenting with our third party provider and will be working with them to improve responses;
For some responses, it would be expected to receive null

GuillaumeRx: I would rather make it optional and not return anything if the value is `null`

SIPS/sip-29.md
```diff
@@ -243,13 +269,138 @@ type FungibleAssetMetadata = {
...
+type NonFungibleAssetMarketData = {
+  // Represents a non-fungible asset market data.
+  fungible: false;
+
+  // The last sale of one asset in the collection.
+  lastSale?: {
+    // The asset that was sold.
+    asset: Caip19AssetTypeOrId;
+    // The price at which is was sold represented as a decimal number in a string.
+    amount: string;
```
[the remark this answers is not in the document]
GuillaumeRx: I don't see the usefulness of having those two values 😅 They both represent the same thing

SIPS/sip-29.md
```diff
@@ -243,13 +269,138 @@ type FungibleAssetMetadata = {
...
+### Asset Market Data
+```typescript
+type NonFungibleAssetMarketData = {
+  // Represents a non-fungible asset market data.
+  fungible: false;
+
+  // The last sale of one asset in the collection.
+  lastSale?: {
+    // The asset that was sold.
+    asset: Caip19AssetTypeOrId;
```
sahar-fehri, line 368: I see the asset returned by the nft-api looks like this
`{
            "type": "native",
            "name": "Wrapped SOL",
            "symbol": "SOL",
            "decimals": 9,
            "token_id": "So11111111111111111111111111111111111111112"
          }`
What would the snap return in this case?

SIPS/sip-29.md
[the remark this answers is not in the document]
GuillaumeRx: Same answer ! 😂 `2900000000 === 2.9` it's just a different format !

SIPS/sip-29.md
GuillaumeRx: The snap would use its assets data to get the SOL CAIP-19 ID and return that. The data that the API returns is not CAIP compliant so we need to do an extra step to convert it :) Overall the idea is not just do dump bind our API to the Snap API as it's clearly not the intention of the Snaps APIs but the snap itself can deal with transforming the data to the expected shape

SIPS/sip-29.md
sahar-fehri: Sounds good thank you 🙏

SIPS/sip-29.md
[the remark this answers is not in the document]
sahar-fehri: lool sorry for some reason i haven't seen your reply before posting the second question; alrighty thnx :D

SIPS/sip-29.md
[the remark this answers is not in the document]
sahar-fehri: Hmmm why would it not be an NFT anymore? 
I'm thinking this is same as ERC1155 for EVMs where you could hold a balance > 1 of the same NFT;
I think the more specific term would be "SFT"; example of a collection is this https://magiceden.io/marketplace/genopets_terraform_seeds_sft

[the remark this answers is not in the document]
GuillaumeRx: @danroc what's your take on this case ?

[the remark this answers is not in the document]
GuillaumeRx: @sahar-fehri does the NFT API already supports those standards ?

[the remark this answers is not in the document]
danroc: I’d prefer to leave it out for now unless it’s already supported by the NFT API. If the backend doesn’t support it yet, we can add it as an optional field later.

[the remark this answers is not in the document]
sahar-fehri: The NFT-API is returning the balance property. From what i tested, the third party providers is not including the SFT in the response.
What i was thinking is to have the snap also return the "balance" property so the UI is able to read it directly from state.
If you guys prefer to leave out the "balance" property; The client might need to hardcode the balance value as "1" instead of reading it from state.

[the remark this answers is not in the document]
GuillaumeRx: Where do we read the balance in the UI ?

SIPS/sip-29.md
```diff
@@ -243,13 +269,144 @@ type FungibleAssetMetadata = {
...
+
+  // Represents a non-fungible asset
+  fungible: false;
+
+  // The time at which the asset was acquired.
+  // The time is represented as a UNIX timestamp. 
+  acquiredAt?: number;
+
+  // Indicates whether the asset is a spam asset.
+  isSpam?: boolean;
```
FrederikBolding, line 355: Why would a Snap ever return a spam asset?

GuillaumeRx: https://github.com/MetaMask/SIPs/pull/174#discussion_r2095959816

FrederikBolding:
```suggestion
  isPossibleSpam?: boolean;
```
Maybe something like this then so that the Snap can flag it as a possible spam token and the client can deal with it based on that?

SIPS/sip-29.md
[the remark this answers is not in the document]
sahar-fehri: For visibility, the send flow in the UI is using the balance, i'm fine to leave this out for now 🙏

sahar-fehri (approved):

danroc (approved):
LTGM

SIPS/sip-29.md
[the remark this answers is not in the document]
GuillaumeRx: We might do something else if we ever have to return the balance of SFTs, the Asset metadata shouldn't have the balance as it's not really a metadata

PR #176, opened 2025-06-03
SIP-20: WebSockets rewrite

FrederikBolding:
Rewrites SIP-20 with a more narrow WebSockets scope, to be implemented in the very near future.

Closes https://github.com/MetaMask/MetaMask-planning/issues/3665

### DIFF
--- a/SIPS/sip-20.md
+++ b/SIPS/sip-20.md
@@ -4,7 +4,7 @@ title: WebSockets
 status: Draft
 author: David Drazic (@david0xd), Frederik Bolding (@frederikbolding), Guillaume Roux (@guillaumerx) 
 created: 2023-12-15
-updated: 2024-06-05
+updated: 2025-06-05
 ---
 
 ## Abstract
--- a/SIPS/sip-20.md
+++ b/SIPS/sip-20.md
@@ -151,7 +151,7 @@ snap.request({
 
 ### Handling WebSocket Events
 
-Snaps can handle WebSocket events by implementing the `onWebSocketEvent` handler. This handler will be called whenever a WebSocket event occurs, such as receiving a message, opening a connection, closing a connection, or encountering an error.
+Snaps can handle WebSocket events by implementing the `onWebSocketEvent` handler. This handler will be called whenever a WebSocket event occurs, such as receiving a message, opening a connection or closing a connection.
 
 ```typescript
 import { OnWebSocketEventHandler } from "@metamask/snap-sdk";
@@ -169,10 +169,6 @@ export const onWebSocketEvent: OnWebSocketEventHandler = async ({ event }) => {
     case "close":
       // Handle connection closed
       console.log(`WebSocket connection closed with ID ${event.id} from ${event.origin}`);
-      break;
-    case "error":
-      // Handle error
-      console.error(`WebSocket error occurred with ID ${event.id} from ${event.origin}`);
       break;
   }
 };
@@ -196,8 +192,7 @@ the type for an `onWebSocketEvent` handler function's arguments.
 export type WebSocketEvent =
   | WebSocketMessage
   | WebSocketOpenEvent
-  | WebSocketCloseEvent
-  | WebSocketErrorEvent;
+  | WebSocketCloseEvent;
 ```
 
 ##### Message Event
@@ -277,24 +272,6 @@ export type WebSocketCloseEvent = {
 `reason` - A string providing a human-readable explanation of why the WebSocket connection was closed.
 
 `wasClean` - A boolean indicating whether the connection was closed cleanly (i.e., without any errors).
-
-##### Error event
-
-This event is triggered when an error occurs with a WebSocket connection. It provides the unique identifier of the connection, and the origin of the event.
-
-```typescript
-export type WebSocketErrorEvent = {
-  type: "error";
-  id: string;
-  origin: string;
-};
-```
-
-`type` - The type of the WebSocket event, which is `"error"` for this event type.
-
-`id` - The unique identifier of the WebSocket connection associated with the event.
-
-`origin` - The origin of the WebSocket connection where the error occurred.
 
 ## Copyright
 
--- a/SIPS/sip-20.md
+++ b/SIPS/sip-20.md
@@ -220,32 +220,24 @@ export type WebSocketEvent =
 This event is triggered when a message is received over a WebSocket connection. The message can be either text or binary data.
 
 ```typescript
-export type WebSocketTextMessage = {
-  type: "message";
-  id: string;
-  origin: string;
+export type WebSocketTextMessageData = {
   dataType: "text";
   data: string;
 };
-```
+export type WebSocketBinaryMessageData = {
+  dataType: "binary";
+  data: number[];
+};
 
-`type` - The type of the WebSocket event, which is `"message"` for this event type.
+export type WebSocketMessageData =
+  | WebSocketTextMessageData
+  | WebSocketBinaryMessageData;
 
-`id` - The unique identifier of the WebSocket connection associated with the event.
-
-`origin` - The origin of the WebSocket event.
-
-`dataType` - The type of data received in the event, which is `"text"` for this event type.
-
-`data` - The data received in the event, which is a string for text messages.
-
-```typescript
-export type WebSocketBinaryMessage = {
+export type WebSocketMessage = {
   type: "message";
   id: string;
   origin: string;
-  dataType: "binary";
-  data: number[];
+  data: WebSocketMessageData;
 };
 ```
 
@@ -255,15 +247,7 @@ export type WebSocketBinaryMessage = {
 
 `origin` - The origin of the WebSocket event.
 
-`dataType` - The type of data received in the event, which is `"binary"` for this event type.
-
-`data` - The data received in the event, which is a number array for binary messages.
-
-```typescript
-export type WebSocketMessageEvent = 
-  | WebSocketTextMessage
-  | WebSocketBinaryMessage;
-```
+`data` - The data received in the event. This can be either a text message (as a string) or binary data (as an array of numbers). The `type` property indicates whether the data is text or binary.
 
 ##### Connection opened event
 
--- a/SIPS/sip-20.md
+++ b/SIPS/sip-20.md
@@ -6,184 +6,46 @@ author: David Drazic (@david0xd)
 created: 2023-12-15
 ---
 
-## Abstract
+- `snap_openConnection`
+- `snap_closeConnection`
+- `snap_sendMessage`
 
-This SIP proposes a new permission that enables Snaps to receive data from external sources.
-
-In the initial version, this SIP proposes support for WebSockets, but it can be extended to support other data sources like blockchain events in the future.
-The Snap can specify the external data source in the Snap manifest. The client then connects to the external data source and sends the data to the Snap.
-The Snap can then do whatever it wants with the data. This initial version only supports one-way communication from the external source to the Snap. The Snap can't send any data back to the external source.
-
-## Motivation
-
-Snaps are currently limited in their ability to receive data from external sources; they have to rely on user actions or cron jobs to fetch data, so they can't react to events in real time. Snaps also cannot use WebSocket connections to receive data from external sources, and are limited to HTTP requests.
-
-## Specification
-
-> Formal specifications are written in Typescript.
-
-### Language
-
-The key words "MUST", "MUST NOT", "SHOULD", "RECOMMENDED" and "MAY" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
-
-### Snap manifest
-
-This SIP introduces a new permission named `endowment:external-data`.
-
-```typescript
-/**
- * A WebSocket connection as specified in the Snap manifest.
- *
- * @property type - The type of connection. Currently, only WebSockets are
- * supported.
- * @property url - The URL to connect to. This must be a valid WebSocket URL,
- * starting with `wss://`. URLs starting with `ws://` are not allowed.
- * @property dataType - The type of data expected from the WebSocket. If this
- * field is omitted, text is assumed.
- */
-type WebSocketConnection = {
-  type: 'websocket';
-  url: string;
-  dataType?: 'text' | 'binary';
+```ts
+export type WebSocketDataEvent = {
+  type: "data";
+  id: string;
+  origin: string;
+  dataType: "text" | "binary";
+  data: string | Uint8Array;
 };
 
-/**
- * An external data connection as specified in the Snap manifest.
- *
- * Currently, only {@link WebSocketConnection} is supported.
- */
-type ExternalDataConnection = WebSocketConnection;
-
-/**
- * External data connections as specified in the Snap manifest.
- *
- * This is the value of the `endowment:external-data` field in the Snap
- * manifest.
- *
- * @property maxRequestTime - The maximum request time for the `onExternalData`
- * entry point, as described in SIP-21.
- * @property connections - The external data connections.
- */
-type ExternalData = {
-    maxRequestTime?: number;
-    connections: ExternalDataConnection[]
+export type WebSocketOpenEvent = {
+  type: "open";
+  id: string;
+  origin: string;
 };
+
+export type WebSocketCloseEvent = {
+  type: "close";
+  id: string;
+  origin: string;
+};
+
+export type WebSocketErrorEvent = {
+  type: "error";
+  id: string;
+  origin: string;
+};
+
+export type WebSocketEvent =
+  | WebSocketDataEvent
+  | WebSocketOpenEvent
+  | WebSocketCloseEvent
+  | WebSocketErrorEvent;
 ```
 
-#### Example
-
-The new field can be specified as follows in a `snap.manifest.json` file:
-
-```json
-{
-  "initialPermissions": {
-    "endowment:external-data": {
-      "maxRequestTime": 50000,
-      "connections": [
-        {
-          "type": "websocket",
-          "url": "wss://example.com/binary/endpoint",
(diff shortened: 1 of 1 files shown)


SIPS/sip-20.md
```diff
@@ -2,188 +2,213 @@
...
 
-export const onExternalData: OnExternalDataHandler = ({ type, data, source }) => {
-  assert(type === 'websocket'); // `data` is inferred as `WebSocketData`.
-  assert(data.type === 'text'); // `message` is inferred as `string`.
+`id` - The unique identifier of the WebSocket connection associated with the event.
 
-  // Now the Snap can do whatever is needed with the message.
-  const json = JSON.parse(data.message);
-};
+`origin` - The origin of the Snap that is handling the WebSocket event.
```
FrederikBolding, line 202:
```suggestion
`origin` - The origin of the WebSocket event.
```

SIPS/sip-20.md
```diff
@@ -2,188 +2,213 @@
...
 
-The `url` MUST start with `wss://` which is a protocol indicator for secure WebSocket connection.
+- `id` - The unique identifier of the WebSocket connection.
+- `url` - The URL of the WebSocket connection.
 
-### Snap implementation
+```typescript
+type WebSocketConnection = {
+  id: string;
+  url: string;
```
FrederikBolding, line 110: We should choose either URL or origin and use that everywhere

SIPS/sip-20.md
```diff
@@ -2,188 +2,213 @@
 sip: 20
 title: External data entry point
```
FrederikBolding, line 3: WebSockets

FrederikBolding, line 3:
```suggestion
title: WebSockets
```

SIPS/sip-20.md
GuillaumeRx: I agree that for open connections `origin` is the way to go but to open a connection like in `snap_openWebSocket` we should use `url` as `origin` doesn't make sense

SIPS/sip-20.md
```diff
@@ -2,188 +2,213 @@
...
+});
 ```
+#### `snap_sendMessage`
+This method allows a Snap to send a message over an existing WebSocket connection. The method takes the unique identifier of the connection and the message to send as parameters.
 
-### Permission caveats
+```typescript
+type SendWebSocketMessageParams = {
+  id: string;
+  message: string | Uint8Array;
```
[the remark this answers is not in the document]
GuillaumeRx: Yeah let's start here and we can always update later

SIPS/sip-20.md
```diff
@@ -1,189 +1,237 @@
...
+
+- Any incoming message from WebSocket connections MUST be handled by the client and forwarded to the appropriate Snap through `onWebSocketEvent`. 
+
+- When the Snap requests to open a WebSocket connection via `snap_openWebSocket`, the client MUST ensure that the connection is established and return a unique identifier for that connection.
+
+- Any open connection MUST be kept alive by the client until the Snap explicitly closes it or the connection is lost, even when the Snap is terminated.
+
+- When the Snap requests to close a WebSocket connection via `snap_closeWebSocket`, the client MUST close the connection and remove it from its list of open connections.
+
+- Any errors that occur during the WebSocket connection lifecycle (e.g., connection failure, message send failure) MUST be handled by the client and reported to the Snap via `onWebSocketEvent`.
```
FrederikBolding, line 41: I'm wondering if this applies to everything. Should we for instance not throw on `snap_openWebSocket` if the initial connection fails? 🤔

SIPS/sip-20.md
```diff
@@ -1,189 +1,237 @@
...
-The key words "MUST", "MUST NOT", "SHOULD", "RECOMMENDED" and "MAY" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" written in uppercase in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Top-level architecture
+
+This architecture defers WebSocket management to the client, allowing Snaps to open, close, and send messages over WebSocket. The client will handle the underlying WebSocket connections and notify the Snap of any events that occur on those connections.
+
+The client has multiple responsibilities in this architecture:
```
Mrtenz, line 30:
```suggestion
The client has multiple responsibilities in this architecture:

```

SIPS/sip-20.md
```diff
@@ -1,189 +1,237 @@
...
+
+```typescript
+snap.request({
+  method: "snap_closeWebSocket",
+  params: {
+    id: "unique-connection-id",
+  },
+});
 ```
+#### `snap_sendMessage`
```
Mrtenz, line 99: I think this name is too generic, especially given that everything else is specific to WebSockets.

SIPS/sip-20.md
```diff
@@ -1,189 +1,237 @@
...
+```typescript
+type SendWebSocketMessageParams = {
+  id: string;
+  message: string | number[];
+};
+```
+
+The RPC method takes two parameters:
+- `id` - The unique identifier of the WebSocket connection to send the message over. This identifier is returned by the `snap_openWebSocket` method.
+- `message` - The message to send over the WebSocket connection. It can be either a string or a `Uint8Array`.
```
Mrtenz, line 111: This doesn't match the type above.

SIPS/sip-20.md
```diff
@@ -1,189 +1,237 @@
...
-type WebSocketBinaryMessage = {
-  type: 'binary';
-  message: ArrayBuffer;
+```typescript
+export type WebSocketMessageEvent = {
+  type: "message";
+  id: string;
+  origin: string;
+  dataType: "text" | "binary";
+  data: string | Uint8Array;
```
Mrtenz, line 184: Ideally we use a discriminated union here to determine the type of `data`, like in the previous specification.

SIPS/sip-20.md
```diff
@@ -1,189 +1,237 @@
...
+`id` - The unique identifier of the WebSocket connection associated with the event.
 
-  // Now the Snap can do whatever is needed with the message.
-  const json = JSON.parse(data.message);
-};
+`origin` - The origin of the WebSocket event.
+
+`dataType` - The type of data received in the event, which can be either `text` or `binary`. This property is only present for `message` events.
+
+`data` - The data received in the event. For `message` events, this can be either a string (for text messages) or a `Uint8Array` (for binary messages). For other event types, this property is not present.
```
Mrtenz, line 229: These shouldn't be documented here if they aren't present. Seems more clear to describe the different events separately.

SIPS/sip-20.md
```diff
@@ -1,189 +1,237 @@
...
+
+This handler does not return any value. It is used to handle WebSocket events asynchronously:
```
Mrtenz, line 231:
```suggestion
This handler does not return any value.
```

SIPS/sip-20.md
```diff
@@ -1,191 +1,298 @@
...
-The Snap can then do whatever it wants with the data. This initial version only supports one-way communication from the external source to the Snap. The Snap can't send any data back to the external source.
+This SIP proposes to expose a new communication protocol to `endowment:network-access`, that enables Snaps to communicate with external services via WebSockets. This will allow Snaps to receive real-time data updates from external sources, such as price feeds or event notifications.
 
 ## Motivation
-
-Snaps are currently limited in their ability to receive data from external sources; they have to rely on user actions or cron jobs to fetch data, so they can't react to events in real time. Snaps also cannot use WebSocket connections to receive data from external sources, and are limited to HTTP requests.
+Currently, Snaps can only communicate with external services via HTTP requests. This limits their ability to receive real-time data updates, which is essential for many use cases, such as price feeds or event notifications. By exposing a WebSocket protocol, Snaps can establish persistent connections with external services and receive real-time updates.
```
FrederikBolding:
```suggestion
Currently, Snaps can only communicate with external services via HTTP requests. This limits their ability to receive real-time data updates, which is essential for many use cases, such as price feeds or event notifications. By exposing the WebSocket protocol, Snaps can establish persistent connections with external services and receive real-time updates.
```

SIPS/sip-20.md
```diff
@@ -1,191 +1,313 @@
...
-  assert(type === 'websocket'); // `data` is inferred as `WebSocketData`.
-  assert(data.type === 'text'); // `message` is inferred as `string`.
+##### Error event
 
-  // Now the Snap can do whatever is needed with the message.
-  const json = JSON.parse(data.message);
+This event is triggered when an error occurs with a WebSocket connection. It provides the unique identifier of the connection, and the origin of the event.
+
+```typescript
+export type WebSocketErrorEvent = {
```
Mrtenz, line 283: This should probably include the error? Or does the WebSocket API not give an actual error message?

SIPS/sip-20.md
```diff
@@ -1,191 +1,298 @@
...
+};
+```
 
-Each caveat object MUST have `type` property that SHOULD be `websocket`.
-The caveat MUST have `url` and `dataType` properties.
-The `url` MUST be string.
-The `dataType` MUST be string that MUST be `"binary"` or `"text"`.
+The RPC method takes two parameters:
+- `id` - The unique identifier of the WebSocket connection to send the message over. This identifier is returned by the `snap_openWebSocket` method.
+- `message` - The message to send over the WebSocket connection. It can be either a string or a number array.
```
FrederikBolding:
```suggestion
- `message` - The message to send over the WebSocket connection. It can be either a string or an array of bytes.
```
Perhaps?

SIPS/sip-20.md
GuillaumeRx: The `error` event don't give any error message, the event type for errors is a generic `Event`. You get a code and a reason in the `close` event, that's why I just added these properties to the close event

SIPS/sip-20.md
```diff
@@ -1,191 +1,298 @@
...
 
-The `url` MUST start with `wss://` which is a protocol indicator for secure WebSocket connection.
+- `id` - The unique identifier of the WebSocket connection.
+- `origin` - The origin of the WebSocket connection.
 
-### Snap implementation
+```typescript
+type WebSocketConnection = {
+  id: string;
+  origin: string;
```
FrederikBolding: Should this be URL?

SIPS/sip-20.md
```diff
@@ -1,191 +1,298 @@
...
+This event is triggered when a message is received over a WebSocket connection. The message can be either text or binary data.
+
+```typescript
+export type WebSocketTextMessage = {
+  type: "text";
+  message: string;
+};
+export type WebSocketBinaryMessage = {
+  type: "binary";
+  message: number[];
```
Mrtenz, line 214: Wondering if this should be a base64 string, since we use that in other places for binary data.

SIPS/sip-20.md
GuillaumeRx: https://stackoverflow.com/a/18804298

SIPS/sip-20.md
GuillaumeRx: @FrederikBolding

SIPS/sip-20.md
FrederikBolding: Do we need to emit this event at all then?

FrederikBolding: It might not be useful if it provides no information at all 😅

SIPS/sip-20.md
FrederikBolding: What would be the reasoning? When I proposed this I was thinking we wanted to mirror the browser WebSocket API as much as possible

SIPS/sip-20.md
FrederikBolding: Hmm, I was assuming an error would always close, but maybe not 🤷‍♂️

GuillaumeRx: This would allow us to be close to the og. WS

SIPS/sip-20.md
Mrtenz: Just consistency with other APIs. Even better if we could have a translation layer on the execution side turning it into `Uint8Array` before passing to the Snap, but 🤷

SIPS/sip-20.md
```diff
@@ -1,191 +1,301 @@
 ---
 sip: 20
-title: External data entry point
+title: WebSockets
 status: Draft
-author: David Drazic (@david0xd)
+author: David Drazic (@david0xd), Frederik Bolding (@frederikbolding), Guillaume Roux (@guillaumerx) 
 created: 2023-12-15
+updated: 2024-06-05
```
danroc, line 7:
```suggestion
updated: 2025-06-05
```

SIPS/sip-20.md
```diff
@@ -1,191 +1,301 @@
...
-  dataType?: 'text' | 'binary';
+  protocols?: string[];
 };
+```
+The RPC method takes one parameter:
+
+- `url` - The URL of the WebSocket service to connect to. 
+  - The URL MUST be a valid WebSocket URL, starting with `wss://`. URLs starting with `ws://` are not allowed.
+  - Only one WebSocket connection can be opened per URL at a time. If a Snap tries to open a new connection to the same URL while an existing connection is still open, the client MUST throw an error.
+- `protocols` - An optional array of subprotocols to use for the WebSocket connection.
```
danroc, line 67: What is the usage for this field?

FrederikBolding: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket#protocols

SIPS/sip-20.md
```diff
@@ -1,191 +1,301 @@
...
-This SIP introduces a new permission named `endowment:external-data`.
+### Snap Manifest
+
+This SIP doesn't introduce any new permissions, but rather extends `endowment:network-access` with new capabilities.
+
+### RPC Methods
+
+#### `snap_openWebSocket`
+
+This method allows a Snap to open a WebSocket connection to an external service. The method takes a URL as a parameter and returns a unique identifier for the connection.
```
danroc, line 53: Can the WebSocket be used immediately after calling `snap_openWebSocket`, or do we need to wait for the `"open"` event?

SIPS/sip-20.md
```diff
@@ -1,191 +1,301 @@
...
+  id: string;
+  origin: string;
+};
 ```
+`type` - The type of the WebSocket event, which is `"open"` for this event type.
 
-#### Example
+`id` - The unique identifier of the WebSocket connection associated with the event.
+
+`origin` - The origin of the WebSocket connection.
```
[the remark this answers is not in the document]
FrederikBolding: The WebSocket origin for easy access to what origin you have open, similar to https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/origin

SIPS/sip-20.md
```diff
@@ -1,191 +1,301 @@
...
+
+##### Error event
+
+This event is triggered when an error occurs with a WebSocket connection. It provides the unique identifier of the connection, and the origin of the event.
+
+```typescript
+export type WebSocketErrorEvent = {
+  type: "error";
+  id: string;
+  origin: string;
```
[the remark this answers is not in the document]
GuillaumeRx: Yes but a snap might keep in state the id <-> origin correlation so it might be easier for it to check the origin

SIPS/sip-20.md
danroc: Just a note: if, for some reason, the Snap needs to use two sub-protocols, it will need to open two connections:

> Note however that only one sub-protocol can be selected per connection.

SIPS/sip-20.md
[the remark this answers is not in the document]
danroc: It's clear in the case of a `MessageEvent`:

> ... origin of the message emitter.

But in the case of a `WebSocketOpenEvent` event, the `origin` is the Snap. Is it correct?

Mrtenz (changes requested):

SIPS/sip-20.md
```diff
@@ -1,191 +1,278 @@
...
+snap.request({
+  method: "snap_sendWebSocketMessage",
+  params: {
+    id: "unique-connection-id",
+    message: "Hello, WebSocket!",
+  },
+});
+```
+#### `snap_getWebSockets`
+This method allows a Snap to retrieve a list of all currently open WebSocket connections. It returns an array of objects, each containing the unique identifier and URL of the connection.
```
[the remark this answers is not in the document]
Mrtenz: The description is outdated now.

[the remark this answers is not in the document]
FrederikBolding:
```suggestion
This method allows a Snap to retrieve a list of all currently open WebSocket connections. It returns an array of objects, each containing the unique identifier, the optional protocols and the URL of the connection.
```

SIPS/sip-20.md
```diff
@@ -1,191 +1,280 @@
...
+    message: "Hello, WebSocket!",
+  },
+});
+```
+#### `snap_getWebSockets`
+This method allows a Snap to retrieve a list of all currently open WebSocket connections. It returns an array of objects, each containing the unique identifier, the optional protocols and the URL of the connection.
 
-The `url` MUST start with `wss://` which is a protocol indicator for secure WebSocket connection.
+- `id` - The unique identifier of the WebSocket connection.
+- `url` - The URL of the WebSocket connection.
```
Mrtenz, line 132:
```suggestion
- `url` - The URL of the WebSocket connection.
- `protocols` - The optional protocols of the WebSocket connection.
```

Mrtenz (approved):

FrederikBolding:
LGTM but I cannot approve my own PR 😅

